Issue #17, opened 2019-09-27
Provide few types of default `Kaspresso` implementations


dector:
Currently `Kaspresso.default()` is feature-loaded (e.g. contains a screenshot interceptor for each step).

As discussed, default implementation should be more lightweight and we can provide few more reconfigured implementations for different hypothetical environments (e.g. local, CI etc).

dector:
I've checked all components in `Kaspresso.default()` implementation and seems that almost all of them are relatively "lightweight" except screenshot components: `ScreenshotStepWatcherInterceptor` and `TestRunnerScreenshotWatcherInterceptor`.

I suggest making `Kaspresso.default()` simple without screenshots and `Kaspresso.fullFeatured()` (requires better name) with full-featured testing environment..

matzuk:
In master

PR #38, opened 2019-10-29
For #17. Split `default()` environment to `simple()` and `advanced()`.

dector:
_For the purpose of API discussion. Tests will be added when API will be approved._

`Kaspresso` now offers two pre-configured implementations:
- `.simple()` - with logging and flaky-safe behavior;
- `.advanced()` - same as `simple` with additional screenshot features.

`Kaspresso.default()` is deprecated as it's just a synonymous for `.advanced()` (to avoid introducing breaking changes).


matzuk:
I propose to implement this pr like the part of https://github.com/KasperskyLab/Kaspresso/issues/49

PR #51, opened 2019-11-19
Fix getException method

PStrelchenko:
It should not throw exception, only get single exception from list.


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/internal/extensions/other/ThrowableExt.kt
```diff
@@ -32,8 +35,8 @@ internal inline fun <reified ERROR : Throwable, LISTENER> Iterable<LISTENER>.for
 
 internal fun <T : Throwable> List<T>.getException(): Throwable? {
     return when (this.size) {
-        1 -> throw this[0]
```
RuslanMingaliev: 👍

PR #52, opened 2019-11-20
Allow to disable screenshot timestamps

aartikov:


matzuk (changes requested):
Hi! Thank you for your contribution!
The first question here is why do you need to remove timestamps? Timestamp resolves the following problem. If your test passes more than one time per concrete suite then you will have only screenshots from the last attempt (each new attempt overwrites old one). That's why the timestamp in the test's name comes to help.
But. We can suggest another solution. Put the test's screenshots in a new folder each time. I think it will be more predictable way. What do you think?

aartikov:
Hi. I plan to use [VisualReview](https://github.com/xebia/VisualReview) for visual regression analysis. This tool matches screenshots by names and calculates differences. So I need to send a bunch of screenshots with determinated names after test runs.
May be, separate folders is a better choice. But still there should be a reliable way to grab required folder.

matzuk:
Hi @aartikov!
I suggest completing this PR =)
There is a concern here. You resolve local task breaking some logic (saving of screenshots of all tests running per suite). 
Do you want to expand your solution and implement the logic where screenshots will be saved in different folders after each suite but screenshots' names will be without timestamps?
You can take some minds and code from DocLoc solution.

aartikov:
@matzuk no problem. Yes, I can try.

PR #53, opened 2019-11-22
Feature/improve system dialog interceptor

matzuk:


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/systemsafety/SystemDialogSafetyProviderImpl.kt
```diff
@@ -17,8 +16,13 @@ class SystemDialogSafetyProviderImpl(
...
 ) : SystemDialogSafetyProvider {
 
+    companion object {
+        private const val DEFAULT_TIMEOUT: Long = 2000
+    }
+
     private val attemptsToSuppress: List<(UiDevice) -> Unit> = listOf(
-        { uiDevice -> uiDevice.findObject(UiSelector().resourceId("android:id/button1")).click() },
+        { uiDevice -> uiDevice.wait(Until.findObject(By.res("android:id/button1")), DEFAULT_TIMEOUT).click() },
+        { uiDevice -> uiDevice.wait(Until.findObject(By.textContains("Close")), DEFAULT_TIMEOUT).click() },
```
RuslanMingaliev: Получается какой-то language specific. Нет какого-то ресурса системного для этого текста? Вроде `android.R.string.close`?

qwert2603: By the way, there may be autofill popup. It would be usefull to close it, too.

matzuk: Есть вот только такое

    <string name="close_button_text">Close</string>

matzuk: Но в принципе, это лучше, согласен

matzuk: > By the way, there may be autofill popup. It would be usefull to close it, too.

Do you have ideas about how to close such dialog?

eakurnikov: Outside the autofilled app you could only tap outside to close this dialog

qwert2603: @matzuk @eakurnikov we can check that these is autocomplete window using
`InstrumentationRegistry.getInstrumentation().uiAutomation.windows`:

```
if (windows.any { it.isFocused && it.isActive && it.parent != null && it.title == "Параметры автозаполнения" }) {
    Timber.d("closing autofill")
    UiDevice.getInstance(InstrumentationRegistry.getInstrumentation()).pressBack()
}
```

Weak thing about it is title's string. May be, there is string-res.

matzuk: So, we have lambda with ```pressBack``` here. The task is resolved =)

qwert2603: > So, we have lambda with `pressBack` here. The task is resolved =)

But `SystemDialogSafetyProviderImpl.isAndroidSystemDetected` will not detect autocomplete dialog.

eakurnikov: I think the best approach is to disable the autofill service before the test starts. It is better then making this interceptor more complex.
just call context.getSystemService(AutofillManager::class.java).disableAutofillServices()

PR #66, opened 2019-12-27
Make flakysafety default exceptions available outside Kaspresso

AndreySBer:
Need to add additional exceptions to flakySafely for us as:
```kotlin
flakySafely(allowedExceptions = DEFAULT_ALLOWED_EXCEPTIONS + AssertionFailedError::class.java) {
     //doSmth
}
```


matzuk:
@PStrelchenko what do you think? I remember that you asked to keep MutableSet instead of Set for a more suitable way to update ```allowedExceptions```.

PStrelchenko:
@matzuk I think, that we should keep mutable variables only in `Kaspresso.Builder` and it's properties, so it's ok to keep `MutableSet` in `FlakySafetyParams`. But in `Kaspresso` instance, in all methods we should use immutable params, for the common good =) From this point of view I like @AndreySBer 's solution.

@AndreySBer Thanks for your PR! One question for you - why you don't just add your exception in `Kaspresso.Builder`? Do you really need to add your exception into `flakySafely` only in the one single case or it's ok to add this exception for all cases? 

```
class MyTestCase : TestCase(
    kaspressoBuilder = Kaspresso.Builder.default().apply {
        flakySafetyParams.allowedExceptions = mutableSetOf<Class<out Throwable>>().apply { 
            this += flakySafetyParams.allowedExceptions
            this += AssertionFailedError::class.java
        }
    }
)
```

Alarm, the following section is very-very subjective =)

@AndreySBer I don't really like your way for providing extension of `allowedExceptions` in `flakySafely` method =/

```
@Test
fun myTest() {
	init { ... }
		.run {
			flakySafely(allowedExceptions = DEFAULT_ALLOWED_EXCEPTIONS + MyException::class.java) {
				...
			}
		}
}
``` 

When you use static variable `DEFAULT_ALLOWED_EXCEPTIONS` you don't clearly express context of this variable (espessially if you use static import). For me, it would be better if you could access `flakySafetyParams` of your `TestCase` through `Kaspresso` instance. 

```
@Test
fun myTest() {
	init { ... }
		.run {
			flakySafely(allowedExceptions = flakySafetyParams.allowedExceptions + MyException::class.java) {
				...
			}
		}

}
``` 


Why is this better, than `DEFAULT_ALLOWED_EXCEPTIONS`:

a) You have context for `allowedExceptions` 
b) Your static constants remain inside proper class and do not expose for everyone
c) If you need access to other fields of `flakySafetyParams` - you have it

=)

You can expose this params through `TestAssistantsProvider` interface, like this:

```
interface TestAssistantsProvider {

    ...
    val flakySafetyParams: FlakySafetyParams

}
```

RuslanMingaliev:
Fully agree with @PStrelchenko 's opinion about static variable. The rest here is good:)

matzuk:
Yep, I like Pavel's approach. Suggest to implement it =)
@AndreySBer

AndreySBer:
@PStrelchenko together with @matzuk we extracted `kaspresso.params` to `BaseTestContext`.
I don't fully understand, which variable should I make mutable back?

matzuk:
@PStrelchenko Паш, тебе норм?

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/testcases/core/testcontext/BaseTestContext.kt
```diff
@@ -26,4 +26,7 @@ open class BaseTestContext internal constructor(
     ContinuouslyProvider by ContinuouslyProviderImpl(kaspresso.params.continuouslyParams, kaspresso.libLogger),
     ComposeProvider by ComposeProviderImpl(kaspresso),
     WebComposeProvider by WebComposeProviderImpl(kaspresso),
-    TestAssistantsProvider by TestAssistantsProviderImpl(kaspresso)
\ No newline at end of file
+    TestAssistantsProvider by TestAssistantsProviderImpl(kaspresso) {
+
+    val params = kaspresso.params
```
PStrelchenko: This should be in `TestAssistantsProviderImpl`, not here.

PStrelchenko:
> @PStrelchenko Паш, тебе норм?

С иммутабельностью - ок, с пробросом параметров - не очень. Отписал коммент.

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/testcases/core/testcontext/BaseTestContext.kt
matzuk: I think it's a more appropriate way.

AndreySBer: Fixed

PR #67, opened 2019-12-29
Feature/main section enricher

PStrelchenko:
Introduce new feature - **main section enrichers** 

**Motivation** 

We want to have ability to add some functionality for all tests in single place, maybe some `step`s, before and after `run` block of our test cases. Also, we want to use our custom data from`TestContext`, when we write our functionality. 

**Why we can't use existing solutions?**

For now, we have only one mechanism for extending tests functionality -- `TestRunWatcherInterceptor`.  But this class is not generic. It has no information about real data types that we can use in `BaseTestCase`. It has `BaseTestContext`, but this class also doesn't have any information about `Data` type.

**How to use:**

To use enrichers, we define our own `BaseTestCase` implementation to clarify our data types and all enrichers in single place:

```
abstract class MyBaseTestCase<MyTestCaseDsl, MyTestCaseData>(
    kaspresso = Kaspresso.Builder.default(),
    dataProducer  = { action -> MyDataProducer.initData(action) },
    mainSectionEnrichers = listOf(
        object : MainSectionEnricher<MyTestCaseData> {
            override fun beforeMainSectionRun(testInfo: TestInfo, testContext: TestContext<MyTestCaseData>) {
                // do something
            }
        }
    )
)
```

In `do something` block you can do whatever you want with `testContext`, for example, add new steps that will be added into total report, our use `Scenario`s. Also, you have access to all data generated by your `dataProducer` in `init` block of your test case:

```
with(testContext) {
    step("Step 1") { ... }
    step("Step 2") { ... }
    scenario(SilentLogin(data))
}
```


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/testcases/core/TestRunner.kt
```diff
@@ -127,11 +137,14 @@ internal class TestRunner<InitData, Data>(
         checkTestInfo(testInfo)
 
         try {
+            val testContext = TestContext(kaspresso, stepsManager, data)
             testRunWatcherInterceptor.onMainSectionStarted(testInfo)
 
+            mainSectionEnricher.beforeMainSectionRun(testInfo, testContext)
```
matzuk: I think it would be safer to send a copy of ```testInfo``` (but not the original instance).

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/testcases/core/TestRunner.kt
```diff
@@ -127,11 +137,14 @@ internal class TestRunner<InitData, Data>(
...
             steps.invoke(
-                TestContext(kaspresso, stepsManager, data)
+                testContext
             )
+            mainSectionEnricher.afterMainSectionRun(testInfo, testContext)
```
matzuk: I think it would be safer to send a copy of ```testInfo``` (but not the original instance).

sample/src/androidTest/java/com/kaspersky/kaspressample/configurator_tests/enricher_tests/enrichers/AssertionMainSectionEnricher.kt
```diff
@@ -0,0 +1,19 @@
+package com.kaspersky.kaspressample.configurator_tests.enricher_tests.enrichers
+
+import com.kaspersky.kaspressample.configurator_tests.enricher_tests.data.EnricherTestData
+import com.kaspersky.kaspresso.enricher.MainSectionEnricher
+import com.kaspersky.kaspresso.testcases.core.testcontext.TestContext
+import com.kaspersky.kaspresso.testcases.models.info.TestInfo
+
+class AssertionMainSectionEnricher : MainSectionEnricher<EnricherTestData> {
+
+    override fun afterMainSectionRun(testInfo: TestInfo, testContext: TestContext<EnricherTestData>) {
```
matzuk: It'a not a DSL approach =(
I propose the following way to express.
Dirty draft:
```
class MainSectionEnricherImpl<Data> : MainSectionEnricher<Data> {

    private var beforeAction: (TestContext<Data>.(TestInfo) -> Unit)? = null

    fun beforeMainSectionRun(testLambda: TestContext<Data>.(TestInfo) -> Unit) {
        beforeAction = testLambda
    }

    override fun beforeMainSectionRun(testInfo: TestInfo, testContext: TestContext<Data>) {
        beforeAction?.invoke(testContext, testInfo)
    }

}

class SomeUsage {

    fun test() {
        val enricher: MainSectionEnricherImpl<Unit> = MainSectionEnricherImpl()
        enricher.beforeMainSectionRun { testInfo ->
            step("some step") {
                testInfo.testName
                // bla-bla-bla
            }
        }
    }

}
```
As a user, I want to use beforeMainSectionRun in DSL-style. So, I think, we have to provide for users a slightly different class (not interface MainSectionEnricher) to offer a possibility of writing in so sweet manner =)

sample/src/androidTest/java/com/kaspersky/kaspressample/configurator_tests/enricher_tests/enrichers/LoggingMainSectionEnricher.kt
```diff
@@ -0,0 +1,19 @@
+package com.kaspersky.kaspressample.configurator_tests.enricher_tests.enrichers
+
+import com.kaspersky.kaspressample.configurator_tests.enricher_tests.data.EnricherTestData
+import com.kaspersky.kaspresso.enricher.MainSectionEnricher
+import com.kaspersky.kaspresso.testcases.core.testcontext.TestContext
+import com.kaspersky.kaspresso.testcases.models.info.TestInfo
+
```
matzuk: I think it's this feature is an ideal candidate to be presented in the documentations =)

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/testcases/core/TestRunner.kt
PStrelchenko: Hmm, `TestInfo` already is data class with immutable fields. Do we really need to provide a copy?

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/testcases/core/TestRunner.kt
PStrelchenko: Hmm, `TestInfo` already is data class with immutable fields. Do we really need to provide a copy?

sample/src/androidTest/java/com/kaspersky/kaspressample/configurator_tests/enricher_tests/enrichers/AssertionMainSectionEnricher.kt
PStrelchenko: Ok, we've discussed this proposal more detailed. 
We want to add DSL approach into enrichers implementations, like this:

```
class MyEnricher : BaseMainSectionEnricher<Data>() {

     override fun enrichMainSection() {
          beforeMainSectionRun { testInfo ->
              // here,  this == TestContext<Data> , so you can add steps in the following way
              step("My step in before run") { ... }
              step("My step 2 in before run") { ... }
          }

          afterMainSectionRun { testInfo -> 
              //  this == TestContext<Data> 
          }
     }

}
```

sample/src/androidTest/java/com/kaspersky/kaspressample/configurator_tests/enricher_tests/enrichers/LoggingMainSectionEnricher.kt
PStrelchenko: Ok, I'll add documentation into `Kaspresso configurator.md`

PR #69, opened 2020-01-05
[DO NOT MERGE] Snapshots support

RuslanMingaliev:
Added https://oss.jfrog.org/artifactory snapshots 

Need to configure CircleCI and Webhooks before merge


RuslanMingaliev:
Is not actual

PR #70, opened 2020-01-05
[DRAFT]Safe extensions for Assertions. BaseAssertions

matzuk:


matzuk:
A lot of developers are against this approach because they are afraid of using such a conception in a non-suitable way.

PR #71, opened 2020-01-08
Set configurator when enabling accessibility services

darran-kelinske-fivestars:
This is really just for convenience. If you don't set this, then when accessing a UiDevice, the accessibility service will be killed.


matzuk:
Thanks, guys!
@dazza5000 please sign CLA =)

matzuk:
@dazza5000 
I think that we can set ```FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES``` in Kaspresso file because a developer may use UiAutomator through not only AccessibilityImpl that's why the mentioned bug will stay here.
Can you improve your solution and replace your code into ```Kaspresso.Builder``` (where we initialize ```uiDevice``` property)?

dazza5000:
k - are you saying to implement it when retrieving the UiDevice from kaspresso?

PR #76, opened 2020-01-16
feature=Kautomator_views: BottomNavigationView + Checkbox

RuslanMingaliev:
BottomNavigationView and Checkbox implementations for Kautomator


kautomator/src/main/kotlin/com/kaspersky/components/kautomator/dsl/check/UiCheckableAssertions.kt
```diff
@@ -23,24 +24,8 @@ interface UiCheckableAssertions : UiBaseAssertions {
         view.check(CheckableAssertionType.IS_NOT_CHECKED) { assertThat(isChecked).isFalse() }
     }
 
-    /**
-     * Checks if the view is checkable
-     */
-    fun isCheckable() {
```
RuslanMingaliev: Removed this method due to their uselessness. Checkable actions should be applied to views that implement Checkable interface.

PR #226, opened 2021-02-10
Removed private constructor to allow Params customization

Starmel:


Starmel:
@matzuk

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/params/AutoScrollParams.kt
```diff
@@ -9,7 +9,7 @@ import junit.framework.AssertionFailedError
  * The class that holds all the necessary for [com.kaspersky.kaspresso.autoscroll.AutoScrollProviderImpl] and
  * [com.kaspersky.kaspresso.autoscroll.WebAutoScrollProviderImpl] parameters.
  */
-class AutoScrollParams private constructor(
+class AutoScrollParams(
```
matzuk, line 12: Hi @Starmel!
Thank you for your contribution!
Observing other such classes (`ContinuouslyParams.kt`, `FlakySafetyParams.kt`), I suggest adding a factory method instead of making the constructor public.

matzuk (changes requested):

matzuk:
done in #290

PR #240, opened 2021-04-19
Updated AutoScrollProviderImpl to allow scrolling on NestedScrollView #239

### LINKED ISSUE #239: Improvement - Allow AutoScrollProviderImpl to perform scroll on NestedScrollView
Currently, the implementation of `AutoScrollProviderImpl.scroll(...)` is:

```
    @Throws(Throwable::class)
    override fun <T> scroll(interaction: ViewInteraction, action: () -> T, cachedError: Throwable): T {
        return try {
            interaction.perform(ViewActions.scrollTo())
            logger.i("View autoScroll successfully performed.")
            action.invoke()
        } catch (error: Throwable) {
            throw cachedError
        }
    }
```

Which uses `ViewActions.scrollTo()` to perform the scrolling. By default, `ViewActions.scrollTo()` doesn't support `NestedScrollView`, but Kakao has it's own `ViewAction` that supports it, `NestedScrollToAction()`.

Simply replacing  `ViewActions.scrollTo()` with `NestedScrollToAction()` should allow  `AutoScrollProviderImpl` to keep it's current behavior and add support to `NestedScrollView`s.

AndreThlckr:
Closes #239

### DIFF
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/autoscroll/AutoScrollProviderImpl.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/autoscroll/AutoScrollProviderImpl.kt
@@ -1,7 +1,6 @@
 package com.kaspersky.kaspresso.autoscroll
 
 import androidx.test.espresso.ViewInteraction
-import androidx.test.espresso.action.ViewActions
 import com.agoda.kakao.common.actions.NestedScrollToAction
 import com.kaspersky.kaspresso.internal.extensions.other.isAllowed
 import com.kaspersky.kaspresso.logger.UiTestLogger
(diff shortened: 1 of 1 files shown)


matzuk:
Hi Andre!
Many thanks for your contribution! We appreciate any help and especially a contribution =)
@eakurnikov what do you think? Is it safe? Maybe there is a mind to introduce a separate AutoScrollInterceptor? Also, I want to highlight the fact that ```NestedScrollToAction``` was written almost a year ago, that's why we couldn't know about such possibility in Kakao when we had created such interceptors.
@AndreThlckr have you checked your solutions on our samples?

RuslanMingaliev:
LGTM

AndreThlckr:
Hi, @matzuk. Do you mean the tests in the samples? If yes then yeah, they seem to be working just as fine, and I don't think this change would break any corner case or something like that. This NestedScrollToAction is almost a copy of scrollTo, it just adds new item to the internal matcher of allowed parents.

eakurnikov (approved):

PR #256, opened 2021-06-03
Kaspresso allure support

eakurnikov:
After conflicts resolved


RuslanMingaliev:
@eakurnikov перенес изменения сюда: #265

PR #286, opened 2021-10-05
Issue/242-3

matzuk:
Hi @sergio-sastre!
Again, thank you very much for your contribution. You've done a great job!
So, I spent some time thinking about how to make Robololectric support more seamless for users. I've decided to evolve your solution in the following things:
1. A user doesn't need to add `sharedTest = true`. All tests can be run on the JVM environment and Instrumental environment simultaneously without additional configurations (except Robolectric configuration sure). With the previous solution, reusage of the test on both environments was a little bit trickier.
2. `KautomatorConfig` is transformed into `InstrumentalDepsAssistant` and related classes. It allows to reduce boilerplate code and leave some `Device` classes and methods working even on the JVM environment.
3. A user can know and use the environment state through `isInstrumentalEnvironment` property in `TestContext` and `Kaspresso.Builder` classes.
4. All manipulations with inteceptors are living inside `Kaspresso` class. But, a user can do similar things in `Kaspresso.Builder` too.


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/instrumental/exception/DocLocInUnitTestException.kt
```diff
@@ -0,0 +1,10 @@
+package com.kaspersky.kaspresso.instrumental.exception
+
+import java.lang.RuntimeException
+
+class DocLocInUnitTestException : RuntimeException(
+    """
```
matzuk, line 6: add empty rows in the beginning and in the end

wiki/08_Kaspresso-Robolectric.md
```diff
@@ -0,0 +1,150 @@
...
+        run {
+            step("Press Home button") {
+                if (isInstrumentalEnvironment) {
+                    device.exploit.pressHome() // execute this section only on the Instrumental environment
+                }
+            }
+        }
+}
+
+// isInstrumentalEnvironment field is available in Kaspresso.Builder and TestContext.
```
matzuk, line 127: field -> property

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/Device.kt
```diff
@@ -119,22 +120,25 @@ data class Device(
...
-    val context: Context
-        get() = InstrumentationRegistry.getInstrumentation().context
+    val context: Context = instrumentation.context
 
     /**
      * A not caching property to get target [Context].
      */
-    val targetContext: Context
-        get() = InstrumentationRegistry.getInstrumentation().targetContext
+    val targetContext: Context = instrumentation.targetContext
```
RuslanMingaliev, line 137: ?

RuslanMingaliev: > A not caching property to get target [Context].

Is caching now for unknown reason

samples/kaspresso-sample/src/sharedTest/kotlin/com/kaspersky/kaspressample/sharedtest/FailingKautomatorSharedTest.kt
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+    @get:Rule
+    val activityTestRule = ActivityTestRule(MainActivity::class.java, true, false)
+
+    @Test
+    fun instrumentalTest() = kautomatorSampleTest()
+
+    @Test(expected = KautomatorInUnitTestException::class)
+    fun unitTest() = kautomatorSampleTest()
```
RuslanMingaliev, line 32: If the test fails due to environment, it cannot be a part of sharedTest

kaspresso/build.gradle.kts
```diff
@@ -15,13 +15,6 @@ dependencies {
     api(libs.androidXCore)
     api(libs.androidXRules)
 
-    api(projects.kautomator)
```
RuslanMingaliev, line 18: ???

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/instrumental/InstrumentalDepsAssistant.kt
```diff
@@ -0,0 +1,16 @@
...
+import androidx.annotation.RequiresApi
+import androidx.test.uiautomator.UiDevice
+
+interface InstrumentalDepsAssistant {
+
+    val isInstrumentalEnvironment: Boolean
+    val uiDevice: UiDevice
+    val uiAutomation: UiAutomation
+
+    @RequiresApi(Build.VERSION_CODES.N)
```
sergio-sastre, line 14: I believe UiAutomation is available form JELLY_BEAN, API level 16.
Is there any other reason for this?

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/instrumental/InstrumentalDepsAssistantImpl.kt
```diff
@@ -0,0 +1,39 @@
...
+    override val isInstrumentalEnvironment: Boolean
+        get() = checkInstrumentalStateOfEnvironment()
+
+    override val uiDevice: UiDevice
+        get() =
+            if (isInstrumentalEnvironment) UiDevice.getInstance(instrumentation)
+            else throw NotSupportedInstrumentalTestException(location, "UiDevice")
+
+    @RequiresApi(Build.VERSION_CODES.N)
+    override fun getUiAutomation(flags: Int): UiAutomation {
```
sergio-sastre, line 24: same here:
JELLY_BEAN, API level 16

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/instrumental/InstrumentalUtils.kt
```diff
@@ -0,0 +1,13 @@
+package com.kaspersky.kaspresso.instrumental
+
+import androidx.test.platform.app.InstrumentationRegistry
+
+/**
+ * @return true if an environment is instrumental, or false if the environment is JVM (with Robolectric)
+ */
+internal fun checkInstrumentalStateOfEnvironment(): Boolean =
```
sergio-sastre, line 8: I actually prefer your approach to the one `AndroidJunit4.java` follow.

For documentation purposes, I leave it here as a comment, since this is how Google made Espresso tests compatible with robolectric:
```kotlin
private static String getRunnerClassName() {
    String runnerClassName = System.getProperty("android.junit.runner", null);
    if (runnerClassName == null) {
      if (!System.getProperty("java.runtime.name").toLowerCase().contains("android")
          && hasClass("org.robolectric.RobolectricTestRunner")) {
        return "org.robolectric.RobolectricTestRunner";
      } else {
        return "androidx.test.internal.runner.junit4.AndroidJUnit4ClassRunner";
      }
    }
    return runnerClassName;
  }
```

That's why in order to make test able to run on the JVM, one has to use `@RunWith(AndroidJunit4.class)`
You can findAndroidJunit4.class code [here](https://github.com/android/android-test/blob/master/ext/junit/java/androidx/test/ext/junit/runners/AndroidJUnit4.java)

samples/kaspresso-sample/src/sharedTest/kotlin/com/kaspersky/kaspressample/sharedtest/FailingKautomatorSharedTest.kt
```diff
@@ -27,31 +26,30 @@ class FailingKautomatorSharedTest : TestCase(Kaspresso.Builder.simple(sharedTest
...
+                        click()
                     }
                 }
-                step("Push start button") {
-                    UiContinuouslyScreen {
-                        startButton {
-                            click()
-                        }
+            }
+            step("Push start button") {
```
sergio-sastre, line 47: I realized that when running tests that use a **counter or timer to show a view**, they do fail on the JVM. In this case, the `FlakySafeViewBehaviorInterceptor` does not seem to help. I think this is something on the Robolectric side, although they [improved looper simulation](http://robolectric.org/blog/2019/06/04/paused-looper/) for this kind of stuff, but I didn't manage to get it working :(
I believe I deleted this step because of it. It could be that with the new approach

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/instrumental/InstrumentalUtils.kt
```diff
@@ -0,0 +1,13 @@
...
+    try {
+        InstrumentationRegistry.getInstrumentation().uiAutomation != null
+    } catch(exception: Exception) {
```
sergio-sastre, line 11: Maybe better to catch only a more precise Exception, like `NullPointerException`?
In doing so, we can avoid swallowing other undesired exceptions.

kautomator/src/main/kotlin/com/kaspersky/components/kautomator/common/InstrumentalUtils.kt
```diff
@@ -0,0 +1,13 @@
...
+import androidx.test.platform.app.InstrumentationRegistry
+
+/**
+ * @return true if an environment is instrumental, or false if the environment is JVM (with Robolectric)
+ */
+internal fun checkInstrumentalStateOfEnvironment(): Boolean =
+    try {
+        InstrumentationRegistry.getInstrumentation().uiAutomation != null
+    } catch(exception: Exception) {
+        false
```
sergio-sastre, line 12: Same here, maybe better to decrease the range of exceptions caught

sergio-sastre (changes requested):
@matzuk 
Hi there, I got some time to review it! Thanks for the changes!
I like the approach in general, and agree on the sharedTest flag not being needed.

There are 2 points to review:
- @RequiresApi(Build.VERSION_CODES.N) for uiAutomation -> unless there is any other reason, it should be @RequiresApi(Build.VERSION_CODES.JELLY_BEAN_MR2), API 18 as stated in the [docu](https://developer.android.com/reference/android/app/UiAutomation)
-  ActivitiesImpl takes instrumentation as argument. Not sure whether it is allowed on the JVM but it could crash in that case?

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/activities/ActivitiesImpl.kt
```diff
@@ -14,7 +14,8 @@ import org.junit.Assert
  * The implementation of the [Activities] interface.
  */
 class ActivitiesImpl(
-    private val logger: UiTestLogger
+    private val logger: UiTestLogger,
+    private val instrumentation: Instrumentation,
```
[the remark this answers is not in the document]
matzuk: I was wondered but `Instrumentation` works on the JVM environment correctly. Sure, I've not checked the entire class, but `context`, `targetContext` properties of `Instrumentation` are correct, methods like `runOnMainSync` at `ActivitiesImpl` work too. That's why I decided to keep `Instrumentation` outside `InstrumentalDepsAssistant`.
Honestly, I've not checked robolectric sources to understand how they do it.

samples/kaspresso-sample/src/sharedTest/kotlin/com/kaspersky/kaspressample/sharedtest/FailingKautomatorSharedTest.kt
sergio-sastre: I believe it was something I missed when testing it, but I'm glad it really works :)

PR #308, opened 2021-11-07
Fix Robolectric version for Robolectric wiki

utzcoz:
I think the version you want to mention is 4.5.1 instead of 5.4.1, because Robolectric doesn't reach 5.x now.


matzuk:
> I think the version you want to mention is 4.5.1 instead of 5.4.1, because Robolectric doesn't reach 5.x now.

Yep, sure. Thanks!

matzuk (approved):

PR #324, opened 2021-11-22
Issue 310/compose module

matzuk:


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/extensions/ThrowableExt.kt
```diff
@@ -1,9 +1,9 @@
-package com.kaspersky.kaspresso.internal.extensions.other
+package com.kaspersky.kaspresso.extensions
 
 import com.kaspersky.kaspresso.internal.exceptions.KaspressoError
 import io.reactivex.exceptions.ExtCompositeException
 
-internal inline fun <reified T : Throwable> invokeSafely(
```
RuslanMingaliev, line 6: Could we hide this from users somehow? Maybe with reflection or a common module?

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/interceptors/watcher/testcase/impl/composite/TestRunCompositeWatcherInterceptor.kt
```diff
@@ -53,7 +53,7 @@ class TestRunCompositeWatcherInterceptor(
                 .isEmpty()
         ) {
             logger.e(
-                """Please, revert back DefaultTestRunWatcherInterceptor to 
+                """Please, revert back DefaultTestRunWatcherInterceptor to
```
RuslanMingaliev, line 56:
```suggestion
                """Revert back DefaultTestRunWatcherInterceptor to
```

RuslanMingaliev (dismissed):

RuslanMingaliev (approved):

PR #326, opened 2021-11-23
Issue 315/fix kamulator tests

matzuk:


samples/kautomator-sample/src/androidTest/java/com/kaspersky/kaspresso/kautomatorsample/screen/ComponentsScreen.kt
```diff
@@ -7,7 +7,7 @@ import com.kaspersky.components.kautomator.component.dialog.UiAlertDialog
 import com.kaspersky.components.kautomator.component.text.UiButton
 import com.kaspersky.components.kautomator.screen.UiScreen
 
-object ComponentsScreen : UiScreen<ComponentsScreen>() {
+class ComponentsScreen : UiScreen<ComponentsScreen>() {
```
RuslanMingaliev, line 10: We use 3 different ways to work with Page Objects in our tests. I prefer to use objects because it's very simple and clear. I propose you fix the problem for objects rather than using classes, especially with `onUiScreen` which creates a new instance on every call.

matzuk: It makes sense. Fixed

RuslanMingaliev (approved):

PR #328, opened 2021-11-24
#325 minsdk = 18

matzuk:


build-logic/android/src/main/kotlin/convention.android-app.gradle.kts
```diff
@@ -9,6 +9,10 @@ plugins {
 android {
     testBuildType = "debug"
 
+    defaultConfig {
+        multiDexEnabled = true
```
RuslanMingaliev, line 13: This solution doesn't solve the problem. We should decrease the method count in the framework (e.g. use `inline`s). Can discuss it later.

RuslanMingaliev (approved):

build-logic/android/src/main/kotlin/convention.android-app.gradle.kts
matzuk: Agree. Not sure that `inline` rescues =)

PR #321, opened 2021-11-21
Autoscroll error: scrolling to views inside a scrollableView with padding with Kaspresso (works with Kautomator)

sergio-sastre:
I wonder whether the padding issue also happens with **recyclerViews**, but if it does, extra code is needed to handle that case.

Apart from that, feel free to start reviewing :)

PR #344, opened 2021-12-23
Issue/342 kautomator not working properly on views with several scrollables

### LINKED ISSUE #342: Kautomator auto-scrolling not working properly in screens with several Scrollables
This happens for example in screens containing several Scrollable views like one "CollapsingToolbarLayout" and a "ScrollView/NestedScrollView/RecyclerView..."
e.g. the layout I was using to fix ScrollView with padding:
[Layout with CollapsingToolbarLayout, HorizontalScrollableView and NestedScrollView](https://github.com/KasperskyLab/Kaspresso/pull/321/files#diff-6e17f780a56327fff5ff7cec07e92330f978fe48c3b26bebd32f2bb83435c4b0)

Removing the CollapsingToolbarLayout fixes the issue indeed.

That happens because ObjectAutoScrollProvider uses 
```
val scrollable = UiScrollable(UiSelector().scrollable(true))
```

That scrolls the first scrollable View, in this cases the CollapsingToolbarLayout!

Possible solution:
Enable through Kaspresso builder to pass a UiSelector() for the View we want to autoscroll
e.g.
```
class ObjectAutoScrollProviderImpl(
    private val logger: UiTestLogger,
    private val autoScrollParams: AutoScrollParams,
    private val uiScrollSelector: UiSelector = UiSelector().scrollable(true)
) : AutoScrollProvider<UiObjectInteraction> {
...
   val scrollable = UiScrollable(uiScrollSelector)
...

```
This would give the developer the flexibility for such cases.

Any other suggestion to a better solution is welcome :)

NOTE:
I tested a draft of the proposed solution and it works. I can take this also for 1.4.1 :)

Checklist:

- [ ] Works with HorizontalScrollView
- [ ] Works with any "vertical" scrollable view like NestedScrollView

sergio-sastre:
This PR is built on top of the issue-306/flaky_autoscroll_when_scrollview_has_padding

It adds a UiHorizontalScrollView to allow automator to explicitly scroll on HorizontalScrollViews, in order to workaround those cases in which Kautomator autoscroll do not work (i.e. numerous scrollable views on the screen)

**Remark**
I'd rather merge this into issue-306 before merging into master (cannot set target branch to issue-306 though), or rather merge this into master and close the PR for the issue-306, because the code is also in this PR that aims to resolve the scrolling issues with Kautomator and Kaspresso

Closes #342


sergio-sastre (commented):
I believe we should also add sth abut the scrolling issues with Kaspresso (i.e. padding) and Kautomator (i.e. numerous scrollable views in the screen under test) in the readme, so that users that encounter cut problems are aware of how to workaround such problems

eakurnikov (commented):
Could you please change the target pr branch from master to that one? It would be easier to see the difference.
As far as I can see here it is ok.

sergio-sastre:
> Could you please change the target pr branch from master to that one? It would be easier to see the difference.
> As far as I can see here it is ok.

Looks like I cannot really change the master branch

> Could you please change the target pr branch from master to that one? It would be easier to see the difference.
> As far as I can see here it is ok.


> Could you please change the target pr branch from master to that one? It would be easier to see the difference.
> As far as I can see here it is ok.

Hi @eakurnikov !
It seems I cannot set the target branch to `sergio-sastre:issue-306-...`, likely because it is from a forked repo.
So I did it in the forked repo for easier reviewing
https://github.com/sergio-sastre/Kaspresso/pull/2
hope it helps :)

eakurnikov:
@sergio-sastre hey, I've just checked it, it is ok

sergio-sastre:
> @sergio-sastre hey, I've just checked it in your fork, thanks, it is ok. But we need small fixes in the included #321 to merge it

Sure, I'll take a look next week. Thanks for reviewing @eakurnikov !

sergio-sastre:
> @sergio-sastre hey, I've just checked it in your fork, thanks, it is ok. But we need small fixes in the included #321 to merge it

@eakurnikov, @matzuk 
So I was playing around with Kautomator, and I found a way to actually solve the autoscroll for Kautomator for real, going through all the scrollables UIAutomator finds on the screen :)

eakurnikov:
Hey Sergio, thank you a lot, sorry for the delay, I will review the changes in the nearest few days

sergio-sastre:
> Hey @sergio-sastre , thank you a lot, sorry for the delay, I will review the changes in the nearest few days

I'd like to do still some small changes.
I've realised there are 3 situations in which it does not scroll:

- [ ] ViewPagers: In general they also need to call UiScrollable#.setAsHorizontalScroll()
- [ ] RecyclerViews with LinearLayoutOrientation.HORIZONTAL: Also need call to UiScrollable#.setAsHorizontalScroll()
- [ ] Horizontal ScrollableViews in the middle of the main vertical scrolling view, not visible before starting to scroll. The changes I added do only work when those views are visible before starting any scrolling, due to the way uiAutomator works.

1 & 2 are easy to fix without affecting scrolling speed significantly. Although I know how to solve 3, I think it might considerably decrease the scrolling speed, also for those layouts not containing horizontal ScrollableViews at all... I need to sleep on it

PR #353, opened 2022-02-15
Fix META-INF exclusions

RuslanMingaliev:


matzuk (approved):

RuslanMingaliev:
Checked with publication to `localMaven()`. Allure and Compose extensions work now.

PR #362, opened 2022-06-10
Add config for publication to MavenCentral publishing

Fullik:
Всем привет!
Добавил возможность заливки новой версии библиотеки через github actions по ручному триггеру.

### DIFF
--- a/.github/workflows/publication_maven.yml
+++ b/.github/workflows/publication_maven.yml
@@ -49,5 +49,11 @@ jobs:
           echo $'\n'kaspresso.pgp.password=$PGP_PASSWORD >> $HOME/.gradle/gradle.properties
           echo $'\n'stableVersion=$STABLE_VERSION >> $HOME/.gradle/gradle.properties
           echo $'\n'snapshotVersion=$SNAPSHOT_VERSION-SNAPSHOT >> $HOME/.gradle/gradle.properties
+      - name: Static analysis
+        run: ./gradlew detektAll lint
+      - name: Tests
+        run: ./gradlew test
+      - name: Assemble build
+        run: ./gradlew assemble assembleAndroidTest
       - name: Publish
         run: ./gradlew publishRelease
--- a/.github/workflows/publication_maven.yml
+++ b/.github/workflows/publication_maven.yml
@@ -7,14 +7,11 @@ on:
         required: true
         type: string
         description: 'Stable version'
-      snapshotVersion:
-        required: true
-        type: string
-        description: 'Snapshot version'
 
 jobs:
   publication:
     runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
     steps:
       - uses: actions/checkout@v2
       - uses: actions/cache@v1
@@ -28,7 +25,13 @@ jobs:
         with:
           distribution: 'adopt'
           java-version: 11
-      - name: Set up environment variables
+      - name: Static analysis
+        run: ./gradlew detektAll lint
+      - name: Tests
+        run: ./gradlew test
+      - name: Assemble build
+        run: ./gradlew assemble assembleAndroidTest
+      - name: Publish
         env:
           OSSRH_USER: ${{ secrets.ossrh_user }}
           OSSRH_PASSWORD: ${{ secrets.ossrh_password }}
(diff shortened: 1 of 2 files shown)


build-logic/publication/src/main/kotlin/convention.sonatype.gradle.kts
```diff
@@ -77,7 +77,7 @@ signing {
         .orNull
     val signingKey = providers.gradleProperty("kaspresso.pgp.key")
         .forUseAtConfigurationTime()
-        .orNull?.toFile()?.readText()
```
AzamatCherchesov, line 80: @Fullik here it was possible to leave work with a file, not text
Perhaps it would be even more convenient.

AzamatCherchesov (dismissed):

.github/workflows/publication_maven.yml
```diff
@@ -0,0 +1,59 @@
...
+          java-version: 11
+      - name: Set up environment variables
+        env:
+          OSSRH_USER: ${{ secrets.ossrh_user }}
+          OSSRH_PASSWORD: ${{ secrets.ossrh_password }}
+          OSSRH_STAGING_PROFILE_ID: ${{ secrets.ossrh_staging_profile_id }}
+          PGP_KEYID: ${{ secrets.pgp_keyid }}
+          PGP_KEY: ${{ secrets.pgp_key }}
+          PGP_PASSWORD: ${{ secrets.pgp_password }}
+          STABLE_VERSION: ${{ github.event.inputs.stableVersion }}
```
RuslanMingaliev, line 39: Project's [gradle.properties](https://github.com/KasperskyLab/Kaspresso/blob/c83189c06594eaea3fdf9e239a2ff33c740ea553/gradle.properties#L9)  should be the source of truth. Why do you want to override it here?

AzamatCherchesov: It seems that local.properties will be a better place

.github/workflows/publication_maven.yml
```diff
@@ -0,0 +1,59 @@
...
+          SNAPSHOT_VERSION: ${{ github.event.inputs.snapshotVersion }}
+        run: |
+          mkdir -p $HOME/.gradle
+          touch $HOME/.gradle/gradle.properties
+          echo $'\n'kaspresso.ossrh.user=$OSSRH_USER >> $HOME/.gradle/gradle.properties
```
RuslanMingaliev, line 44: It's not safe. Do not write sensitive data to the files, your entire secrets might have been compromised

RuslanMingaliev (changes requested):

build-logic/publication/src/main/kotlin/convention.sonatype.gradle.kts
Fullik: We cant pass to github secrets a file, only string. To prevent any security issues with writing key into the file we have to use this property as text.

.github/workflows/publication_maven.yml
Fullik: Changed passing property to gradle task to passing parameters via gradlew command.

.github/workflows/publication_maven.yml
Fullik: Remove version input

.github/workflows/publication_maven.yml
```diff
@@ -0,0 +1,40 @@
...
+          PGP_KEYID: ${{ secrets.pgp_keyid }}
+          PGP_KEY: ${{ secrets.pgp_key }}
+          PGP_PASSWORD: ${{ secrets.pgp_password }}
+        run: >
+          ./gradlew publishRelease -Pkaspresso.ossrh.user=${OSSRH_USER}
+          -Pkaspresso.ossrh.password=${OSSRH_PASSWORD}
+          -Pkaspresso.ossrh.stagingProfileId=${OSSRH_STAGING_PROFILE_ID}
+          -Pkaspresso.pgp.keyid=${PGP_KEYID}
+          -Pkaspresso.pgp.key=${PGP_KEY}
+          -Pkaspresso.pgp.password=${PGP_PASSWORD}
```
RuslanMingaliev, line 40: --no-parallel

RuslanMingaliev (dismissed):

.github/workflows/publication_maven.yml
Fullik: Added

RuslanMingaliev (approved):

PR #368, opened 2022-06-29
Issue 306/fix padding breaks autoscroll interceptor

Nikitae57:
Created a new interceptor which is invoked when standard one fails. It takes into account scrollable's padding and works with NestedScrollView and CoordinatorLayout


sergio-sastre:
There is already a PR to solve this issue: https://github.com/KasperskyLab/Kaspresso/pull/321

Moreover, this PR does not have any UI test that confirms it works if the scrollview has padding

Nikitae57:
Yeah, missed it. Was hoping that "projects" dashboard had a fresh info and forgot to check PRs

PR #383, opened 2022-09-22
added new developers to the list

AzamatCherchesov:


AzamatCherchesov:
@matzuk @RuslanMingaliev @eakurnikov have a look please

build-logic/publication/src/main/kotlin/convention.publication-base.gradle.kts
```diff
@@ -31,6 +31,26 @@ publishing.publications.withType<MavenPublication> {
...
+                id.set("Fullik")
+                name.set("Pogos Azaryan")
+                url.set("https://github.com/Fullik")
+            }
+            developer {
+                id.set("maxsurovtsev")
+                name.set("Maxim Surovtsev")
+                url.set("https://github.com/maxsurovtsev")
+            }
             developer {
```
eakurnikov, line 54: But why is this the order again?) It's not random, not alphabetic, not just adding new members to the list, there is some logic in inserting

eakurnikov: :)

eakurnikov (approved):

Issue #387, opened 2022-09-29
Kaspresso with LeakAssertions


PaladiRka:
I tried combine Kaspresso with LeakAssertions. And some UI test failed by Kaspresso LanguageImpl.cachedActivity (Log below).
```
Application memory leaks were detected:
====================================
HEAP ANALYSIS RESULT
====================================
2 APPLICATION LEAKS

References underlined with "~~~" are likely causes.
Learn more at https://squ.re/leaks.

382528 bytes retained by leaking objects
Signature: f8cb42f36daef658bb5683aa2c527e197f776713
┬───
│ GC Root: Input or output parameters in native code
│
├─ MyTest instance
│    Leaking: UNKNOWN
│    Retaining 389,4 kB in 5372 objects
│    ↓ BaseTestCase.device
│                   ~~~~~~
├─ com.kaspersky.kaspresso.device.Device instance
│    Leaking: UNKNOWN
│    Retaining 109 B in 3 objects
│    context instance of android.app.ContextImpl
│    targetContext instance of android.app.ContextImpl
│    ↓ Device.language
│             ~~~~~~~~
├─ com.kaspersky.kaspresso.device.languages.LanguageImpl instance
│    Leaking: UNKNOWN
│    Retaining 382,6 kB in 5131 objects
│    cachedActivity instance of MyActivity with mDestroyed = true
│    context instance of android.app.ContextImpl
│    ↓ LanguageImpl.cachedActivity
│                   ~~~~~~~~~~~~~~
╰→ MyActivity instance
​     Leaking: YES (ObjectWatcher was watching this because MyActivity received Activity#onDestroy() callback and Activity#mDestroyed is true)
​     Retaining 382,5 kB in 5129 objects
​     key = 065172ba-73f3-4afc-bbb2-21f155580fee
​     watchDurationMillis = 11770
​     retainedDurationMillis = 6768
​     mApplication instance of MyApp
​     mBase instance of androidx.appcompat.view.ContextThemeWrapper
```

And I suggest fix. LanguageImpl subscribe on ActivityLifecycle and cached activity when it created but don't clear activity when it destroyed. 

https://github.com/KasperskyLab/Kaspresso/blob/c3d6d547ec57b9db8c53eafdf75e1142defb00de/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/languages/LanguageImpl.kt#L22-L26

```kotlin
[...]

PR #388, opened 2022-09-29
Clean up cachedActivity after destroyed it.

PaladiRka:
https://github.com/KasperskyLab/Kaspresso/issues/387

### DIFF
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/languages/LanguageImpl.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/languages/LanguageImpl.kt
@@ -23,6 +23,9 @@ class LanguageImpl(
             if (stage == Stage.CREATED) {
                 cachedActivity = activity
             }
+            if ((stage == Stage.DESTROYED) and (cachedActivity == activity)) {
+                cachedActivity = null
+            }
         }
 
     init {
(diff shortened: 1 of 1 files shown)


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/languages/LanguageImpl.kt
```diff
@@ -23,6 +23,9 @@ class LanguageImpl(
             if (stage == Stage.CREATED) {
                 cachedActivity = activity
             }
+            if ((stage == Stage.DESTROYED) and (cachedActivity == activity)) {
```
RuslanMingaliev, line 26: AFAIK `onDestroy` method call is not guaranteed. What about  `Stage.DESTROYED`?  
May be weak reference on `cachedActivity` is more suitable here?

RuslanMingaliev (dismissed):

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/languages/LanguageImpl.kt
PaladiRka: Yes, is better

PaladiRka:
@RuslanMingaliev

RuslanMingaliev (approved):

AzamatCherchesov (approved):

e13mort:
@matzuk @eakurnikov could you review this MR please?

eakurnikov (approved):

PaladiRka:
Can you tell when this fix will released?

AzamatCherchesov:
@PaladiRka this fix will be included in release 1.4.2, which we plan to publish in a few days

Issue #404, opened 2022-11-02
Target 30 support


AzamatCherchesov:
Correct work with storage (migrate from Manage all external storage permission to Scoped storage)

Nikitae57:
Need to be careful with this. Whole kaspresso is built on using traditional file API. Another thing to consider is screen recording on android 10+. Current implementation uses "abd shell screenrecord path/to/save/video/on/device" (I'll call it just screen recorder onwards). 
- screen recorder could write video to /sdcard/ but we can't create any subfolders from kaspresso without MANAGE_EXTERNAL_STORAGE permission
- screen recorder can't record to /data/data/your.package/files directory because of permissions issue

Nikitae57:
Linked to #407

Nikitae57:
We discussed the idea to put all artifacts into Documents folder since it's a public folder and quick experiment showed us that we could use it without restrictions. With this approach we have to keep in mind that allure stores reports to /data/data. Ideally we want to save reports to the same place as other artifacts. We decided that we'll copy reports from /data/data to Documents and cover it with an optional flag in Kaspresso builder

PR #423, opened 2022-11-15
[Draft] #404: Using documents to avoid storage restrictions

Nikitae57:
Also repaired allure video recording

### DIFF
--- a/allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
+++ b/allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
@@ -1,0 +1,126 @@
+package com.kaspersky.components.alluresupport.interceptors.testrun
+
+import android.app.Instrumentation
+import androidx.test.uiautomator.UiDevice
+import com.google.common.io.CharStreams
+import com.kaspersky.kaspresso.files.dirs.DirsProvider
+import com.kaspersky.kaspresso.files.resources.ResourcesDirsProvider
+import com.kaspersky.kaspresso.files.resources.ResourcesRootDirsProvider
+import com.kaspersky.kaspresso.interceptors.watcher.testcase.TestRunWatcherInterceptor
+import com.kaspersky.kaspresso.testcases.models.info.TestInfo
+import org.json.JSONObject
+import java.io.File
+
+private const val ATTACHMENTS_JSON_FIELD = "attachments"
+private const val NAME_JSON_FIELD = "name"
+private const val SOURCE_JSON_FIELD = "source"
+private const val MP4_EXTENSION = "mp4"
+
+/**
+ * Current allure version stores reports to /data/data/your.package.name/files/allure-results.
+ * This interceptor moves them to default artifacts folder after test and replaces mock videos with real ones
+ */
+class MoveReportsInterceptor(
+    private val instrumentation: Instrumentation,
+    private val dirsProvider: DirsProvider,
+    private val resourcesDirsProvider: ResourcesDirsProvider,
+    private val rootDirsProvider: ResourcesRootDirsProvider,
+    private val stateHolder: TestRunStateHolder,
+    private val device: UiDevice
+) : TestRunWatcherInterceptor {
+    override fun onTestFinished(testInfo: TestInfo, success: Boolean) {
+        val allureTargetDir = moveAllureReportToSdCard()
+        removeStubs(allureTargetDir)
+        moveAttachedVideosToProperDirectories(allureTargetDir)
+        cleanUp()
+    }
+
+    /**
+     * Deletes allure report under /data/data/your.package.name/files and empty directories after moving real videos from them
+     */
+    private fun cleanUp() {
+        getOriginalAllureDir().deleteRecursively()
+        val videosDir = dirsProvider.provideNew(rootDirsProvider.videoRootDir)
+        device.executeShellCommand("rm -rf ${videosDir.absolutePath}")
+
+        dirsProvider.provideNew(rootDirsProvider.screenshotsRootDir).deleteRecursively()
+        dirsProvider.provideNew(rootDirsProvider.logcatRootDir).deleteRecursively()
+        dirsProvider.provideNew(rootDirsProvider.viewHierarchy).deleteRecursively()
+    }
+
+    /**
+     * @return allure results dir under /data/data/your.package.name/files
+     */
+    private fun getOriginalAllureDir(): File = instrumentation.targetContext.filesDir.resolve("allure-results")
+
+    /**
+     * Moves allure report from /data/data/your.package.name/files/allure-report to external storage e.g.
+     * /sdcard/Documents/.../allure-results
+     * @return allure target directory on sdcard
+     */
+    private fun moveAllureReportToSdCard(): File {
+        val allureResultsFile = getOriginalAllureDir()
+        val allureTargetDir = dirsProvider.provideNew(rootDirsProvider.allureRootDir)
+        if (!allureResultsFile.exists()) {
+            throw IllegalArgumentException("Unable to move allure results from $allureResultsFile. File not found")
+        }
+        allureResultsFile.copyRecursively(allureTargetDir)
+
+        return allureTargetDir
+    }
+
+    /**
+     * During allure test stubs are attached to report because real videos can't be recorded directly to /data/data.
+     * After kaspresso moves report to sdcard it has to replace stubs with real videos. Videos are moved using adb shell
+     * 👀 https://issuetracker.google.com/issues/258277873
+     * @param allureTargetDir allure directory under /sdcard
+     */
+    private fun removeStubs(allureTargetDir: File) {
+        allureTargetDir.listFiles()?.forEach {
+            if (it.extension.contains(MP4_EXTENSION, ignoreCase = true)) {
+                it.delete()
+            }
+        }
+    }
+
+    /**
+     * Move (and rename) real videos from their original directories to report dir
+     * @param allureTargetDir allure directory under /sdcard
+     */
+    private fun moveAttachedVideosToProperDirectories(allureTargetDir: File) {
+        stateHolder.attachedVideos.forEach {
+            saveAttachedVideo(it, allureTargetDir)
+        }
+    }
+
+    /**
+     * Used for screen recording workaround:
+     * screen cast can't be recorded to /data/data/.../allure-results because of permissions issues (👀 https://issuetracker.google.com/issues/258277873).
+     * We have to save video on /sdcard and in the same time attach a stub to allure report. Before moving report to /sdcard
+     * we need to parse report and save stub file name to rename and move an actual one to allure report dir after moving report to /sdcard. Then we
+     * remove stub videos in /sdcard and /data/data and move real video file to allure report directory to replace stubs
+     * @param attachedVideo video attached to report
+     * @param allureTargetDir allure directory under /sdcard
+     */
+    private fun saveAttachedVideo(attachedVideo: AttachedVideo, allureTargetDir: File) {
+        val allureReportFile = allureTargetDir.resolve("${stateHolder.lastTestCaseUuid ?: ""}-result.json")
+        if (!allureReportFile.exists()) {
+            throw IllegalStateException("Can't attach video to report because the latter not found. Tried path ${allureReportFile.absolutePath}")
+        }
+
+        allureReportFile.inputStream().use {
+            val json = JSONObject(CharStreams.toString(it.reader()))
+            val attachments = json.getJSONArray(ATTACHMENTS_JSON_FIELD)
+            for (i in 0 until attachments.length()) {
+                val attachment = attachments.getJSONObject(i)
+                val attachmentName = attachment.getString(NAME_JSON_FIELD)
+                if (attachmentName.equals(attachedVideo.attachedStubFile.name, ignoreCase = true)) {
+                    val source = attachment.getString(SOURCE_JSON_FIELD) // Attachment real filename
+                    device.executeShellCommand("mv ${attachedVideo.actualFile.absolutePath} ${allureTargetDir.resolve(source)}")
+                    attachedVideo.attachedStubFile.delete()
+                    break
+                }
+            }
+        }
+    }
+}
--- a/allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/TestRunStateHolder.kt
+++ b/allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/TestRunStateHolder.kt
@@ -1,0 +1,34 @@
+package com.kaspersky.components.alluresupport.interceptors.testrun
+
+import java.io.File
+
+/**
+ * Used to store shared state used by multiple interceptors
+ */
+class TestRunStateHolder {
+    private val _attachedVideos = mutableListOf<AttachedVideo>()
+    val attachedVideos: List<AttachedVideo>
+        get() = _attachedVideos.toList()
+
+    var lastTestCaseUuid: String? = null
+
+    fun rememberAttachedVideo(stubFile: File, actualFile: File) {
+        val attachedVideo = AttachedVideo(attachedStubFile = stubFile, actualFile = actualFile)
+        _attachedVideos.add(attachedVideo)
+    }
+}
+
+/**
+ * This model represents allure video attachment. It's used for screen recording workaround
+ */
+data class AttachedVideo(
+    /**
+     * Stub while which has been used to attach a video to report. Should be replaced with a real video file after moving
+     * report to /sdcard
+     */
+    val attachedStubFile: File,
+    /**
+     * Actual screen record file which has been saved into /sdcard. Should be used to replace a stub in allure report
+     */
+    val actualFile: File
+)
--- a/allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/VideoRecordingTestInterceptor.kt
+++ b/allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/VideoRecordingTestInterceptor.kt
@@ -2,11 +2,22 @@ package com.kaspersky.components.alluresupport.interceptors.testrun
 
 import com.kaspersky.components.alluresupport.files.attachVideoToAllureReport
 import com.kaspersky.kaspresso.device.video.Videos
+import com.kaspersky.kaspresso.files.dirs.AllureDirsProvider
 import com.kaspersky.kaspresso.interceptors.watcher.testcase.TestRunWatcherInterceptor
 import com.kaspersky.kaspresso.testcases.models.info.TestInfo
+import io.qameta.allure.kotlin.Allure
+import io.qameta.allure.kotlin.model.Status
 
+/**
+ * Due to screen recorder bug we have to perform a workaround which requires VideoRecordingTestInterceptor
+ * and MoveReportsInterceptor to be the last two interceptors in allure reports. So if you use VideoRecordingTestInterceptor
+ * be sure to use MoveReportsInterceptor too. Otherwise your report will be under /data/data/your.package.name/files/allure-results
+ * and videos under /sdcard
+ */
 class VideoRecordingTestInterceptor(
-    private val videos: Videos
+    private val videos: Videos,
+    private val allureDirsProvider: AllureDirsProvider,
+    private val stateHolder: TestRunStateHolder
 ) : TestRunWatcherInterceptor {
 
     override fun onTestStarted(testInfo: TestInfo) {
@@ -14,6 +25,16 @@ class VideoRecordingTestInterceptor(
     }
 
     override fun onTestFinished(testInfo: TestInfo, success: Boolean) {
-        videos.saveAndApply { attachVideoToAllureReport() }
+        videos.saveAndApply {
+            val stubFile = allureDirsProvider.provideReportVideoAttachmentStub(this)
+            stubFile.attachVideoToAllureReport()
+            val uuid = Allure.lifecycle.getCurrentTestCase() ?: ""
+            Allure.lifecycle.updateTestCase {
+                it.status = if (success) Status.PASSED else Status.FAILED
+            }
+            Allure.lifecycle.stopTestCase(uuid)
+            Allure.lifecycle.writeTestCase(uuid)
+            stateHolder.rememberAttachedVideo(stubFile = stubFile, actualFile = this)
+        }
     }
 }
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/dirs/AllureDirsProvider.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/dirs/AllureDirsProvider.kt
@@ -1,0 +1,55 @@
+package com.kaspersky.kaspresso.files.dirs
+
+import android.annotation.SuppressLint
+import android.app.Instrumentation
+import android.content.Context
+import android.os.Build
+import android.os.Environment
+import androidx.test.uiautomator.UiDevice
+import com.kaspersky.kaspresso.files.resources.ResourcesRootDirsProvider
+import com.kaspersky.kaspresso.internal.extensions.other.createDirIfNeeded
+import com.kaspersky.kaspresso.internal.extensions.other.createFileIfNeeded
+import java.io.File
+
+class AllureDirsProvider(
+    private val instrumentation: Instrumentation,
+    private val resourcesRootDirsProvider: ResourcesRootDirsProvider,
+    device: UiDevice
+) : DefaultDirsProvider(instrumentation, device) {
+
+    @Suppress("DEPRECATION")
+    @SuppressLint("WorldReadableFiles", "ObsoleteSdkInt")
+    override fun provideNew(dest: File): File {
+        if (isVideoDir(dest)) { // screen recorder can't record to /data/data
+            return super.provideNew(dest)
+        }
+
+        val dir: File = when {
+            Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q -> instrumentation.targetContext.applicationContext.filesDir.resolve(dest)
+            Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP -> Environment.getExternalStorageDirectory().resolve(dest)
+            else -> instrumentation.targetContext.applicationContext.getDir(dest.canonicalPath, Context.MODE_WORLD_READABLE)
+        }
+
+        return dir.createDirIfNeeded()
+    }
+
+    /**
+     * Used for allure report video attachment workaround. Creates stub video file in package private directory so allure could attach it to report
+     * @param actualVideoFile vide file saved by screen recorder
+     * @return stub video file under /data/data
+     */
+    fun provideReportVideoAttachmentStub(actualVideoFile: File): File {
+        val defaultVideoDir = super.provideNew(resourcesRootDirsProvider.videoRootDir).absolutePath
+        val targetVideoDir = instrumentation.targetContext.applicationContext.filesDir.resolve(resourcesRootDirsProvider.videoRootDir).absolutePath
+        val targetFilePath = actualVideoFile.absolutePath.replace(defaultVideoDir, targetVideoDir)
+
+        return File(targetFilePath).parentFile!!
+            .createDirIfNeeded()
+            .resolve(actualVideoFile.name)
+            .createFileIfNeeded()
+    }
+
+    private fun isVideoDir(dest: File): Boolean {
+        return dest == resourcesRootDirsProvider.videoRootDir
+    }
+}
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/dirs/DefaultDirsProvider.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/dirs/DefaultDirsProvider.kt
@@ -5,36 +5,54 @@ import android.app.Instrumentation
 import android.content.Context
 import android.os.Build
 import android.os.Environment
+import androidx.test.uiautomator.UiDevice
 import com.kaspersky.kaspresso.internal.extensions.other.createDirIfNeeded
 import java.io.File
 
-class DefaultDirsProvider(
-    private val instrumentation: Instrumentation
+open class DefaultDirsProvider(
+    private val instrumentation: Instrumentation,
+    private val device: UiDevice
 ) : DirsProvider {
     private val clearedDirs = HashSet<File>()
 
     @Suppress("DEPRECATION")
     @SuppressLint("WorldReadableFiles", "ObsoleteSdkInt")
     override fun provideNew(dest: File): File {
-        val dir: File = when {
-            Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q -> instrumentation.targetContext.applicationContext.filesDir.resolve(dest)
-            Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP -> Environment.getExternalStorageDirectory().resolve(dest)
-            else -> instrumentation.targetContext.applicationContext.getDir(dest.canonicalPath, Context.MODE_WORLD_READABLE)
+        val dir = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
+            Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOCUMENTS).resolve(dest)
+        } else {
+            instrumentation.targetContext.applicationContext.getDir(dest.canonicalPath, Context.MODE_WORLD_READABLE)
         }
+
         return dir.createDirIfNeeded()
     }
 
     override fun provideCleared(dest: File): File {
         if (!clearedDirs.contains(dest)) {
-            clearDir(path = dest, inclusive = false)
+            clearDir(dest, inclusive = false)
             clearedDirs.add(dest)
         }
         return dest.createDirIfNeeded()
     }
+    @Suppress("SameParameterValue")
+    private fun clearDir(dest: File, inclusive: Boolean) {
+        clearDirManually(dest, inclusive)
+        if (dest.exists() && dest.list()?.isNotEmpty() == true) {
+            clearDirThroughShell(dest, inclusive)
+        }
+    }
 
-    private fun clearDir(path: File, inclusive: Boolean) {
+    private fun clearDirThroughShell(dest: File, inclusive: Boolean) {
+        if (inclusive) {
+            device.executeShellCommand("rm -r ${dest.absolutePath}")
+        } else {
+            device.executeShellCommand("find ${dest.absolutePath} -type f -delete")
+        }
+    }
+
+    private fun clearDirManually(path: File, inclusive: Boolean) {
         if (path.isDirectory) {
-            path.listFiles()?.forEach { clearDir(path = it, inclusive = true) }
+            path.listFiles()?.forEach { clearDirManually(path = it, inclusive = true) }
         }
         if (inclusive) {
             path.delete()
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/resources/impl/DefaultResourceFilesProvider.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/resources/impl/DefaultResourceFilesProvider.kt
@@ -43,6 +43,7 @@ class DefaultResourceFilesProvider(
             FileExtension.MP4.toString()
         )
         val resourceDir = resourcesDirsProvider.provide(resourcesRootDirsProvider.videoRootDir, subDir)
+            .createDirIfNeeded()
         return resourceDir.createDirIfNeeded()
             .resolve(resFileName)
             .createFileIfNeeded()
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/resources/impl/DefaultResourcesRootDirsProvider.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/resources/impl/DefaultResourcesRootDirsProvider.kt
@@ -8,4 +8,5 @@ class DefaultResourcesRootDirsProvider : ResourcesRootDirsProvider {
     override val screenshotsRootDir = File("screenshots")
     override val videoRootDir = File("video")
     override val viewHierarchy = File("view_hierarchy")
+    override val allureRootDir = File("allure-results")
 }
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/kaspresso/Kaspresso.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/kaspresso/Kaspresso.kt
@@ -5,7 +5,6 @@ import androidx.test.espresso.Espresso
 import androidx.test.espresso.FailureHandler
 import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.uiautomator.Configurator
-import io.github.kakaocup.kakao.Kakao
 import com.kaspersky.adbserver.common.log.logger.LogLevel
 import com.kaspersky.components.kautomator.intercept.interaction.UiDeviceInteraction
 import com.kaspersky.components.kautomator.intercept.interaction.UiObjectInteraction
@@ -64,8 +63,8 @@ import com.kaspersky.kaspresso.files.resources.impl.DefaultResourcesDirNameProvi
 import com.kaspersky.kaspresso.files.resources.impl.DefaultResourcesDirsProvider
 import com.kaspersky.kaspresso.files.resources.impl.DefaultResourcesRootDirsProvider
 import com.kaspersky.kaspresso.idlewaiting.KautomatorWaitForIdleSettings
-import com.kaspersky.kaspresso.instrumental.InstrumentalDependencyProviderFactory
 import com.kaspersky.kaspresso.instrumental.InstrumentalDependencyProvider
+import com.kaspersky.kaspresso.instrumental.InstrumentalDependencyProviderFactory
 import com.kaspersky.kaspresso.interceptors.behavior.DataBehaviorInterceptor
 import com.kaspersky.kaspresso.interceptors.behavior.ViewBehaviorInterceptor
 import com.kaspersky.kaspresso.interceptors.behavior.WebBehaviorInterceptor
@@ -113,15 +112,16 @@ import com.kaspersky.kaspresso.interceptors.watcher.view.impl.logging.LoggingVie
 import com.kaspersky.kaspresso.interceptors.watcher.view.impl.logging.LoggingWebAssertionWatcherInterceptor
 import com.kaspersky.kaspresso.logger.UiTestLogger
 import com.kaspersky.kaspresso.logger.UiTestLoggerImpl
-import com.kaspersky.kaspresso.params.Params
-import com.kaspersky.kaspresso.params.FlakySafetyParams
-import com.kaspersky.kaspresso.params.ContinuouslyParams
 import com.kaspersky.kaspresso.params.AutoScrollParams
+import com.kaspersky.kaspresso.params.ContinuouslyParams
 import com.kaspersky.kaspresso.params.ElementLoaderParams
-import com.kaspersky.kaspresso.params.StepParams
+import com.kaspersky.kaspresso.params.FlakySafetyParams
+import com.kaspersky.kaspresso.params.Params
 import com.kaspersky.kaspresso.params.ScreenshotParams
+import com.kaspersky.kaspresso.params.StepParams
 import com.kaspersky.kaspresso.params.VideoParams
 import com.kaspersky.kaspresso.testcases.core.testcontext.BaseTestContext
+import io.github.kakaocup.kakao.Kakao
 
 /**
  * The storage of all Kaspresso preferences and entities, such as [AdbServer], [Device] and different interceptors.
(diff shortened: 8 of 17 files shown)


allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
```diff
@@ -0,0 +1,124 @@
...
+
+        dirsProvider.provideNew(rootDirsProvider.screenshotsRootDir).deleteRecursively()
+        dirsProvider.provideNew(rootDirsProvider.logcatRootDir).deleteRecursively()
+        dirsProvider.provideNew(rootDirsProvider.viewHierarchy).deleteRecursively()
+    }
+
+    /**
+     * @return allure results dir under /data/data/your.package.name/files
+     */
+    private fun getOriginalAllureDir(): File = instrumentation.targetContext.filesDir.resolve("allure-results")
```
AzamatCherchesov, line 52: nit: "allure-results" can be moved to const val

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
```diff
@@ -0,0 +1,124 @@
...
+            }
+        }
+    }
+
+    /**
+     * Move (and rename) real videos from their original directories to report dir
+     * @param allureTargetDir allure directory under /sdcard
+     */
+    private fun moveAttachedVideosToProperDirectories(allureTargetDir: File) {
+        stateHolder.attachedVideos.forEach {
```
AzamatCherchesov, line 89: To be honest, I don't really like this solution.
It seems unobvious to me. But what if there was no call to stateHolder before? Another interceptor that works with stateHolder may not be connected.
Maybe it makes sense to do it with one interceptor? To ensure both are connected at the same time

AzamatCherchesov (changes requested):

samples/kaspresso-allure-support-sample/src/androidTest/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/AllureTestRunStateHolderTest.kt
```diff
@@ -0,0 +1,175 @@
...
+        val holder = TestRunStateHolder()
+
+        val uuid = UUID.randomUUID()
+        holder.lastTestCaseUuid = uuid.toString()
+
+        assert(holder.lastTestCaseUuid == uuid.toString())
+    }
+}
+
+private const val ATTACHMENT_SOURCE = "3c56a71e-a04c-499e-acfb-5114f01cff27-attachment.mp4"
```
AzamatCherchesov, line 35: unused code

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/dirs/DefaultDirsProvider.kt
```diff
@@ -2,39 +2,65 @@ package com.kaspersky.kaspresso.files.dirs
...
-            clearDir(path = dest, inclusive = false)
+            clearDir(dest, inclusive = false)
             clearedDirs.add(dest)
         }
         return dest.createDirIfNeeded()
     }
+    @Suppress("SameParameterValue")
+    private fun clearDir(dest: File, inclusive: Boolean) {
+        clearDirManually(dest, inclusive)
+        if (dest.exists() && dest.list()?.size == 0 && shouldClearDirThroughShell(dest)) {
```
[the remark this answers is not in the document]
Nikitae57: Android may return empty list when dir contains files if they don't belong to your app

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/resources/impl/DefaultResourceFilesProvider.kt
```diff
@@ -43,6 +43,7 @@ class DefaultResourceFilesProvider(
             FileExtension.MP4.toString()
         )
         val resourceDir = resourcesDirsProvider.provide(resourcesRootDirsProvider.videoRootDir, subDir)
+            .createDirIfNeeded()
```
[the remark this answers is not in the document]
Nikitae57: Actually, it seems to be redundant here. Dirs provider should've created dir at this point
image

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
```diff
@@ -0,0 +1,124 @@
...
+     * @return allure results dir under /data/data/your.package.name/files
+     */
+    private fun getOriginalAllureDir(): File = instrumentation.targetContext.filesDir.resolve("allure-results")
+
+    /**
+     * Moves allure report from /data/data/your.package.name/files/allure-report to external storage e.g.
+     * /sdcard/Documents/.../allure-results
+     * @return allure target directory on sdcard
+     */
+    private fun moveAllureReportToSdCard(): File {
```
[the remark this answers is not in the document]
Nikitae57: Why? Personally I like when file manipulation methods do this like `createDirIfNeeded()`

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
Nikitae57: Good point. I'll make this interceptor remember uuid when test is started, so there would be no need in additional one. Move interceptor need to be a separate one though because it let us move reports even if video was not recorded. In this case we just move report without any videos. If we merge video and move interceptor we won't be able to move reports from /data/data

samples/kaspresso-allure-support-sample/src/androidTest/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/AllureTestRunStateHolderTest.kt
Nikitae57: Fixed

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/kaspresso/Kaspresso.kt
```diff
@@ -667,7 +667,7 @@ data class Kaspresso(
             if (!::libLogger.isInitialized) libLogger = UiTestLoggerImpl(DEFAULT_LIB_LOGGER_TAG)
             if (!::testLogger.isInitialized) testLogger = UiTestLoggerImpl(DEFAULT_TEST_LOGGER_TAG)
 
-            if (!::dirsProvider.isInitialized) dirsProvider = DefaultDirsProvider(instrumentation)
+            if (!::dirsProvider.isInitialized) dirsProvider = DefaultDirsProvider(instrumentation, instrumentalDependencyProviderFactory.getComponentProvider<Kaspresso>(instrumentation))
```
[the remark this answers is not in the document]
Nikitae57: Done

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/dirs/DefaultDirsProvider.kt
```diff
@@ -2,39 +2,65 @@ package com.kaspersky.kaspresso.files.dirs
...
     override fun provideNew(dest: File): File {
-        val dir: File = when {
-            Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q -> instrumentation.targetContext.applicationContext.filesDir.resolve(dest)
-            Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP -> Environment.getExternalStorageDirectory().resolve(dest)
-            else -> instrumentation.targetContext.applicationContext.getDir(dest.canonicalPath, Context.MODE_WORLD_READABLE)
+        val dir = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
+            Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOCUMENTS).resolve(dest)
+        } else {
+            Environment.getExternalStorageDirectory().resolve(dest)
+            // instrumentation.targetContext.applicationContext.getDir(dest.canonicalPath, Context.MODE_WORLD_READABLE)
```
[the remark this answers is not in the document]
Nikitae57: Done

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/dirs/DefaultDirsProvider.kt
```diff
@@ -2,39 +2,65 @@ package com.kaspersky.kaspresso.files.dirs
...
 import android.os.Build
 import android.os.Environment
+import com.kaspersky.kaspresso.instrumental.InstrumentalDependencyProvider
 import com.kaspersky.kaspresso.internal.extensions.other.createDirIfNeeded
 import java.io.File
 
-class DefaultDirsProvider(
-    private val instrumentation: Instrumentation
+open class DefaultDirsProvider(
+    private val instrumentation: Instrumentation,
```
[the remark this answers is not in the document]
Nikitae57: Done

AzamatCherchesov:
Discussed with @Nikitae57 to remove all permission rules in the repository.

AzamatCherchesov (dismissed):

matzuk:
@eakurnikov would be great to give your review
@Nikitae57 great job man, thank you for very detailed comments

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/TestRunStateHolder.kt
```diff
@@ -0,0 +1,32 @@
+package com.kaspersky.components.alluresupport.interceptors.testrun
+
+import java.io.File
+
+/**
+ * Used to store shared state used by multiple interceptors
+ */
+class TestRunStateHolder {
+    private val _attachedVideos = mutableListOf<AttachedVideo>()
+    val attachedVideos: List<AttachedVideo>
```
matzuk, line 10: why not?
```kotlin
val attachedVideos: List<AttachedVideo> = _attachedVideos
```

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/TestRunStateHolder.kt
```diff
@@ -0,0 +1,32 @@
+package com.kaspersky.components.alluresupport.interceptors.testrun
+
+import java.io.File
+
+/**
+ * Used to store shared state used by multiple interceptors
```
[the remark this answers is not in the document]
Nikitae57: Totally. Do you have any suggestions?

matzuk:
Observed the PR again.
1. I hope that I am wrong but there is sense that we reinvite a wheel. Have you seen the [Allure's possibility to write artifacts on the external SD](https://github.com/allure-framework/allure-kotlin#orchestrator-teststorage) and how [marathon](https://github.com/MarathonLabs/marathon) writes all artifacts?

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/TestRunStateHolder.kt
Nikitae57: A matter of taste

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/TestRunStateHolder.kt
```diff
@@ -0,0 +1,32 @@
...
+ */
+class TestRunStateHolder {
```
VladislavSumin, line 8: Please use more correct name, this name to mush general, but this holder stores only video file references

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/VideoRecordingTestInterceptor.kt
```diff
@@ -2,18 +2,39 @@ package com.kaspersky.components.alluresupport.interceptors.testrun
...
     override fun onTestStarted(testInfo: TestInfo) {
         videos.record("Video_${testInfo.testName}")
     }
 
     override fun onTestFinished(testInfo: TestInfo, success: Boolean) {
-        videos.saveAndApply { attachVideoToAllureReport() }
+        videos.saveAndApply {
+            val stubFile = allureDirsProvider.provideReportVideoAttachmentStub(this)
+            stubFile.attachVideoToAllureReport()
+            val uuid = Allure.lifecycle.getCurrentTestCase() ?: ""
```
VladislavSumin, line 31: okey, we can get this UUID after test, but in case below we store UUID in a field before test, why?

gradle/libs.versions.toml
```diff
@@ -5,10 +5,10 @@ espresso = "3.5.0"
 kakao = "3.2.0"
 kakaoCompose = "0.2.0"
 allure = "2.4.0"
-compose = "1.2.1"
+compose = "1.3.1"
```
VladislavSumin, line 8: nit: More clear update libs with separate pull request

samples/kaspresso-sample/build.gradle.kts
```diff
@@ -53,7 +54,9 @@ dependencies {
     testImplementation(libs.androidXTestExtJunit)
     testImplementation(libs.robolectric)
 
-    debugImplementation(libs.androidXTestFragmentTesting)
+    debugImplementation(libs.androidXTestFragmentTesting) {
+        isTransitive = false
```
VladislavSumin, line 58: why? if it necessary, please, add comment why.

VladislavSumin (changes requested):

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
```diff
@@ -0,0 +1,133 @@
...
+import com.google.common.io.CharStreams
+import com.kaspersky.kaspresso.files.dirs.DirsProvider
+import com.kaspersky.kaspresso.files.resources.ResourcesRootDirsProvider
+import com.kaspersky.kaspresso.interceptors.watcher.testcase.TestRunWatcherInterceptor
+import com.kaspersky.kaspresso.testcases.models.info.TestInfo
+import io.qameta.allure.kotlin.Allure
+import org.json.JSONObject
+import java.io.File
+
+private const val ATTACHMENTS_JSON_FIELD = "attachments"
```
eakurnikov, line 14: nitpick:
why not to simply use private companion object? The separated generated class containing these constants seems redundant.

samples/kaspresso-sample/build.gradle.kts
Nikitae57: Latest compose libs bump in master broke sample tests. There was a library conflict, some methods haven't been found in runtime. This fixes the issue

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/TestRunStateHolder.kt
eakurnikov: Will not the new list be instantiated on each call following your approach?

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
Nikitae57: Personally I don't care about generated classes. This way just reads more clearly

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/VideoRecordingTestInterceptor.kt
Nikitae57: I've encountered situation when Allure.lifecycle.getCurrentTestCase() returned null or wrong UUID if there were multiple tests when I call it in onTestFinished

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
```diff
@@ -0,0 +1,133 @@
...
+private const val NAME_JSON_FIELD = "name"
+private const val SOURCE_JSON_FIELD = "source"
+private const val MP4_EXTENSION = "mp4"
+private const val ALLURE_DIR = "allure-results"
+
+/**
+ * Current allure version stores reports to /data/data/your.package.name/files/allure-results.
+ * This interceptor moves them to default artifacts folder after test and replaces mock videos with real ones
+ */
+class MoveReportsInterceptor(
```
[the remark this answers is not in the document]
Nikitae57: Why? We want to let user pick this interceptor in kaspressoBuilder

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/dirs/AllureDirsProvider.kt
```diff
@@ -0,0 +1,55 @@
...
+import android.content.Context
+import android.os.Build
+import android.os.Environment
+import com.kaspersky.kaspresso.files.resources.ResourcesRootDirsProvider
+import com.kaspersky.kaspresso.instrumental.InstrumentalDependencyProvider
+import com.kaspersky.kaspresso.internal.extensions.other.createDirIfNeeded
+import com.kaspersky.kaspresso.internal.extensions.other.createFileIfNeeded
+import java.io.File
+
+class AllureDirsProvider(
```
eakurnikov, line 14: All allure related stuff should be in `allure-support`

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/files/resources/impl/DefaultAllureResourcesRootDirsProvider.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import com.kaspersky.components.alluresupport.files.resources.AllureResourcesRootDirsProvider
+import java.io.File
+
+class DefaultAllureResourcesRootDirsProvider : AllureResourcesRootDirsProvider {
+    override val logcatRootDir = File("logcat")
+    override val screenshotsRootDir = File("screenshots")
+    override val videoRootDir = File("video")
+    override val viewHierarchy = File("view_hierarchy")
+    override val allureRootDir = File("allure-results")
```
eakurnikov, line 11: My concern is that this dir can be changed by the user.

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/files/dirs/AllureDirsProvider.kt
```diff
@@ -0,0 +1,56 @@
...
+
+class AllureDirsProvider(
+    private val instrumentation: Instrumentation,
+    private val resourcesRootDirsProvider: ResourcesRootDirsProvider,
+    instrumentationDependencyProvider: InstrumentalDependencyProvider
+) : DefaultDirsProvider(instrumentationDependencyProvider) {
+
+    @Suppress("DEPRECATION")
+    @SuppressLint("WorldReadableFiles", "ObsoleteSdkInt")
+    override fun provideNew(dest: File): File {
```
eakurnikov, line 23: Maybe it would be better to declare explicitly two methods for two types of dirs providing? And also name them in a way excluding video/non-video words.

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/files/dirs/AllureDirsProvider.kt
```diff
@@ -0,0 +1,56 @@
...
+
+        return dir.createDirIfNeeded()
+    }
+
+    /**
+     * Used for allure report video attachment workaround. Creates stub video file in package private directory so allure could attach it to report
+     * @param actualVideoFile vide file saved by screen recorder
+     * @return stub video file under /data/data
+     */
+    fun provideReportVideoAttachmentStub(actualVideoFile: File): File {
```
eakurnikov, line 42: I think this doesn't belong to dirs provider. Maybe move it to files provder and name it just `provideStubFile`?

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/files/resources/impl/DefaultAllureResourcesRootDirsProvider.kt
eakurnikov: `io.qameta.allure.kotlin.util.PropertiesUtils` is public, you can use it in `allure-support` module. Seems like easy to support this.

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/files/dirs/AllureDirsProvider.kt
Nikitae57: I started it, but it requires much more code changes that I'd like (DirsProvider and ResourcesDirsProvider imeplemtations). I think for now it's better to leave it as it is and do a refactor later.

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/VideoRecordingTestInterceptor.kt
```diff
@@ -1,19 +1,40 @@
...
+            Allure.lifecycle.updateTestCase {
```
eakurnikov, line 32: Seems like these calls to `Allure.lifecycle` are being performed only by `AllureJunit4` and most likely if `AllureRunner` is set by the user. 

What if user have set the `AllureRunner`? I mean we access the other lib's interior lifecycle management manually from here. And we force the test case to stop and to write resources before the allure does so itself. How will this affect the allure's lifecycle?

What if the user have not set the `AllureRunner`? Seems like `startTestCase` is never called but `startTestCase` is.

As far as I understand the intention here is just to link the video file with the test, so can we generate the uuid by our own and manage it separately from allure like we do it in `AllureMapperStepInterceptor`?

wiki/09_Kaspresso-Allure.md
```diff
@@ -136,3 +137,18 @@ Details for failed test:
 ## _Details that you need to know_
 By default, Kaspresso-Allure introduces additional timeouts to assure the correctness of a Video recording as much as possible. To summarize, these timeouts increase a test execution time by 5 seconds.
 You are free to change these values by customizing `videoParams` in `Kaspresso.Builder`. See the example above.
+
+### _How kaspresso attaches videos to allure report_
+
+To support storage restrictions allure saves reports to package provate directory e.g. /data/data/your.package.name/files/allure-report whereas screen recordings are created by calling "screenrecord" utility through adb shell which is unable to
```
eakurnikov, line 143:
```suggestion
To support storage restrictions allure saves reports to package private directory e.g. /data/data/your.package.name/files/allure-report whereas screen recordings are created by calling "screenrecord" utility through adb shell which is unable to
```

wiki/09_Kaspresso-Allure.md
```diff
@@ -136,3 +137,18 @@ Details for failed test:
...
+save video files to package private directories (see this [issue](https://issuetracker.google.com/issues/258277873)).
+Because report created by one user (your package), but videos are created by another one leads to inability to attach videos 
+to report because your package has no rights on vide files.
+
+This fact makes implementing screen recordings attachments rather tedious. Here's kaspresso approach:
+1. During test we attach stub video file to report. This stub is placed under /data/data
```
eakurnikov, line 149:
```suggestion
1. During the test we attach a stub video file to the report. This stub is placed in /data/data
```

wiki/09_Kaspresso-Allure.md
```diff
@@ -136,3 +137,18 @@ Details for failed test:
...
+2. When test is finished we force allure to save report
```
eakurnikov, line 150:
```suggestion
2. Once the test is finished we force allure to save the report 
```

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/VideoRecordingTestInterceptor.kt
Nikitae57: 1. Yes, idea is to make allure write down report so we could parse it and check which UUID it generated for our video file. 
2. Yes, it affects allure's lifecycle - it stops it. I don't think it is a problem either. We interact with allure lifecycle in the end of the test. Each test is run is a separate process, allure doesn't outlive it
3. It would be strange if user wants to launch allure test without configuring allure runner. I don't think It's kaspresso's problem. We could try to detect whether allure runner is used and crash with a verbal message to make user set up runner

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/dirs/AllureDirsProvider.kt
Nikitae57: Done

wiki/09_Kaspresso-Allure.md
```diff
@@ -136,3 +137,18 @@ Details for failed test:
...
+
+To support storage restrictions allure saves reports to package provate directory e.g. /data/data/your.package.name/files/allure-report whereas screen recordings are created by calling "screenrecord" utility through adb shell which is unable to
+save video files to package private directories (see this [issue](https://issuetracker.google.com/issues/258277873)).
+Because report created by one user (your package), but videos are created by another one leads to inability to attach videos 
+to report because your package has no rights on vide files.
+
+This fact makes implementing screen recordings attachments rather tedious. Here's kaspresso approach:
+1. During test we attach stub video file to report. This stub is placed under /data/data
+2. When test is finished we force allure to save report 
+3. Kaspresso parces report and saves file names generated by allure (when you attach file to report allure renames it with
```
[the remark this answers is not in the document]
eakurnikov: Do I understand it right - we take file names generated by allure and use them while saving files by ourselves?

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/resources/impl/DefaultResourcesRootDirsProvider.kt
```diff
@@ -8,4 +8,5 @@ class DefaultResourcesRootDirsProvider : ResourcesRootDirsProvider {
     override val screenshotsRootDir = File("screenshots")
     override val videoRootDir = File("video")
     override val viewHierarchy = File("view_hierarchy")
+    override val allureRootDir = File("allure-results")
```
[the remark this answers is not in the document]
Nikitae57: Done

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/files/resources/impl/DefaultAllureResourcesRootDirsProvider.kt
Nikitae57: Yes, sorry, I'm an idiot. Looks like I've tried to use PropertiesUtils outside of allure support module

wiki/09_Kaspresso-Allure.md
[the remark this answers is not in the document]
Nikitae57: Yes

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/VideoRecordingTestInterceptor.kt
eakurnikov: My bad, it seemed to me `AllureRunner` is optional.

eakurnikov: Tests are not necessary executed each in its own process, there may be no orchestration. Let me check my concern

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
```diff
@@ -0,0 +1,133 @@
...
+    private val dirsProvider: DirsProvider,
+    private val rootDirsProvider: ResourcesRootDirsProvider,
+    private val stateHolder: TestRunStateHolder,
+    private val device: UiDevice
+) : TestRunWatcherInterceptor {
+
+    private var lastTestCaseUuid: String? = null
+
+    override fun onMainSectionStarted(testInfo: TestInfo) {
+        lastTestCaseUuid = Allure.lifecycle.getCurrentTestCase()
```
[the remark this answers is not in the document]
VladislavSumin: ok, when why you can't get `Allure.lifecycle.getCurrentTestCase()` on endSection? you use it inside another interceptor?

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/VideoRecordingTestInterceptor.kt
VladislavSumin: hmmm, i think we need more info about this situation, may be it is allure bug?

samples/kaspresso-sample/build.gradle.kts
VladislavSumin: please add comment here, with error and error reason

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/VideoRecordingTestInterceptor.kt
eakurnikov: `updateTestCase`, `stopTestCase`, `writeTestCase` will be called the second time by allure. And the `updateTestCase` contains little bit different logic for status. Will return on this tomorrow, maybe we could avoid this.

samples/kaspresso-sample/build.gradle.kts
Nikitae57: Done

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
```diff
@@ -0,0 +1,133 @@
...
+    private val instrumentation: Instrumentation,
+    private val dirsProvider: DirsProvider,
+    private val rootDirsProvider: AllureResourcesRootDirsProvider,
+    private val videosHolder: AttachedAllureVideosHolder,
+    private val device: UiDevice
+) : TestRunWatcherInterceptor {
+
+    private var lastTestCaseUuid: String? = null
+
+    override fun onMainSectionStarted(testInfo: TestInfo) {
```
eakurnikov, line 34: Why is it `onMainSectionStarted` and not `onTestStarted` by the way?

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * During allure test stubs are attached to report because real videos can't be recorded directly to /data/data.
+     * After kaspresso moves report to sdcard it has to replace stubs with real videos. Videos are moved using adb shell
+     * 👀 https://issuetracker.google.com/issues/258277873
+     * @param allureTargetDir allure directory under /sdcard
+     */
+    private fun removeStubs(allureTargetDir: File) {
+        allureTargetDir.listFiles()?.forEach {
+            if (it.extension.contains(MP4_EXTENSION, ignoreCase = true)) {
```
eakurnikov, line 87: Maybe we could define the stub more precisely? Not just every mp4 file

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/files/resources/impl/DefaultAllureResourcesRootDirsProvider.kt
```diff
@@ -0,0 +1,14 @@
...
+import com.kaspersky.components.alluresupport.files.resources.AllureResourcesRootDirsProvider
+import io.qameta.allure.kotlin.util.PropertiesUtils
+import java.io.File
+
+class DefaultAllureResourcesRootDirsProvider : AllureResourcesRootDirsProvider {
+    override val logcatRootDir = File("logcat")
+    override val screenshotsRootDir = File("screenshots")
+    override val videoRootDir = File("video")
+    override val viewHierarchy = File("view_hierarchy")
+    override val allureRootDir = File(PropertiesUtils.resultsDirectoryPath)
```
eakurnikov, line 12: Since the `allure.properties` can't be changed in runtime maybe we could cache the `PropertiesUtils.resultsDirectoryPath`'s result and access the cached value in `MoveReportsInterceptor` to not load it on each call?

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
```diff
@@ -0,0 +1,133 @@
...
+) : TestRunWatcherInterceptor {
+
+    private var lastTestCaseUuid: String? = null
+
+    override fun onMainSectionStarted(testInfo: TestInfo) {
+        lastTestCaseUuid = Allure.lifecycle.getCurrentTestCase()
+    }
+
+    override fun onTestFinished(testInfo: TestInfo, success: Boolean) {
+        val allureTargetDir = moveAllureReportToSdCard()
```
eakurnikov, line 39: If the user adds custom interceptors after this one he wouldn't be able to see its results in the report. For example `AllureSupportCustomizeTest` will right custom view hierarchy which will not be added to the report.

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * Deletes allure report under /data/data/your.package.name/files and empty directories after moving real videos from them
+     */
+    private fun cleanUp() {
+        getOriginalAllureDir().deleteRecursively()
+        val videosDir = dirsProvider.provideNew(rootDirsProvider.videoRootDir)
+        device.executeShellCommand("rm -rf ${videosDir.absolutePath}")
+
+        dirsProvider.provideNew(rootDirsProvider.screenshotsRootDir).deleteRecursively()
```
eakurnikov, line 53: This would be a breaking change. Not sure that we have to clear this, previously these files were just copied to the report. I would say we can leave it as it was, but not insisting, need more opinions here. At least it should be mentioned in the changelog.

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/MoveReportsInterceptor.kt
eakurnikov: First I would suggest to move here that hacky forcing the allure to write down the report to have something like:
```
    override fun onTestFinished(testInfo: TestInfo, success: Boolean) {
        forceAllureTestFinished(success)
        val allureTargetDir = moveAllureReportToSdCard()
        removeStubs(allureTargetDir)
        moveAttachedVideosToProperDirectories(allureTargetDir)
        cleanUp()
    }
    fun forceAllureTestFinished(success: Boolean) {
        val uuid = Allure.lifecycle.getCurrentTestCase() ?: “”
        Allure.lifecycle.updateTestCase {
            it.status = if (success) Status.PASSED else Status.FAILED
        }
        Allure.lifecycle.stopTestCase(uuid)
        Allure.lifecycle.writeTestCase(uuid)
    }
```
And then we should guarantee that this `MoveReportsInterceptor` is called last no matter how much interceptors the user added by himself. How to achieve this is need to be designed.

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/interceptors/testrun/VideoRecordingTestInterceptor.kt
eakurnikov: We could but then we need to provide our own instrumentation runner with the run listener. So yeah seems like this hack is the only way. Also there will be errors printed in logcat by allure after each test saying that the testcase can't be found to be stopped.

eakurnikov:
If someone needs more context, pls take a look description and comments [here](https://github.com/KasperskyLab/Kaspresso/pull/438)

samples/kaspresso-sample/src/androidTest/kotlin/com/kaspersky/kaspressample/device_tests/DeviceFilesSampleTest.kt
```diff
@@ -1,17 +1,15 @@
...
-import androidx.test.rule.GrantPermissionRule
 import com.kaspersky.kaspressample.MainActivity
 import com.kaspersky.kaspresso.device.files.Files
 import com.kaspersky.kaspresso.testcases.api.testcase.TestCase
-import java.io.File
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
 import org.junit.Rule
 import org.junit.Test
+import java.io.File
```
AzamatCherchesov, line 12: Why this dependency declaration was moved?

samples/kaspresso-sample/src/androidTest/kotlin/com/kaspersky/kaspressample/device_tests/DeviceLanguageSampleTest.kt
```diff
@@ -1,27 +1,19 @@
...
-import android.Manifest
 import androidx.test.ext.junit.rules.activityScenarioRule
-import androidx.test.rule.GrantPermissionRule
 import com.kaspersky.kaspressample.MainActivity
 import com.kaspersky.kaspressample.screen.MainScreen
 import com.kaspersky.kaspresso.testcases.api.testcase.TestCase
-import java.util.Locale
 import org.junit.Rule
 import org.junit.Test
+import java.util.Locale
```
AzamatCherchesov, line 9: why this dependency declaration was moved?

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/runner/listener/KaspressoLateRunListener.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.runner.Description
+
+interface KaspressoLateRunListener : KaspressoRunListener {
+
+    fun lateInit(cache: Cache) {
+        testRunStarted(cache.testRunStartedDescription)
+        testStarted(cache.firstTestStartedDescription)
+    }
+
+    class Cache {
```
AzamatCherchesov, line 12: Add some info in comments what is it about. Move to notifier class

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/runner/listener/KaspressoRunNotifier.kt
```diff
@@ -0,0 +1,70 @@
...
+        "com.kaspersky.kaspresso.runner.KaspressoRunner is not set"
+    }
+
+internal class KaspressoRunNotifierImpl : KaspressoRunNotifier {
+    private val cache = KaspressoLateRunListener.Cache()
+    override val listeners: MutableList<KaspressoRunListener> = arrayListOf()
+
+    override fun addListener(listener: KaspressoRunListener) {
+        if (listener is KaspressoLateRunListener) {
+            listener.lateInit(cache)
```
AzamatCherchesov, line 36: add some info in comments why do we need this method invocation. (Because some listeners are added after tests started and they left call back invocations)

AzamatCherchesov: Allure listener and allure runner

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/runner/listener/KaspressoLateRunListener.kt
```diff
@@ -0,0 +1,31 @@
+package com.kaspersky.kaspresso.runner.listener
+
+import org.junit.runner.Description
+
+interface KaspressoLateRunListener : KaspressoRunListener {
```
AzamatCherchesov, line 5: why it is late? Add some details

samples/kaspresso-allure-support-sample/src/androidTest/kotlin/com/kaspersky/kaspresso/alluresupport/AllureSupportSanityTest.kt
```diff
@@ -0,0 +1,168 @@
...
+private const val STATUS_JSON_FIELD = "status"
+private const val PASSED = "passed"
+private const val JSON_EXTENSION = "json"
+
+/**
+ * Checks that allure support is sane
+ */
+// TODO implement allure support sanity test in another way. Current code doesn't fail test if report is invalid
+@Ignore("Since we switched to run notifiers this test can no longer be functional")
+class AllureSupportSanityTest : TestCase(
```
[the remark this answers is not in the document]
Nikitae57: Moved to a separate folder

samples/kaspresso-sample/src/androidTest/kotlin/com/kaspersky/kaspressample/docloc_tests/DocLocSanityTest.kt
```diff
@@ -0,0 +1,139 @@
...
+    "1__Simple_screen",
+    "2__Simple_fragment_-_two_buttons",
+    "3__Simple_fragment_-_input",
+    "4__Simple_fragment_-_typed_text",
+)
+
+/**
+ * Check that doc loc tests are sane
+ */
+class DocLocSanityTest : DocLocScreenshotTestCase(
```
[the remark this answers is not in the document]
Nikitae57: Moved to a separate folder

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/runner/KaspressoRunner.kt
```diff
@@ -0,0 +1,24 @@
...
+import androidx.test.runner.AndroidJUnitRunner
+import com.kaspersky.kaspresso.runner.listener.KaspressoRunNotifier
+import com.kaspersky.kaspresso.runner.listener.KaspressoRunNotifierImpl
+import com.kaspersky.kaspresso.runner.listener.SpyRunListener
+
+@Suppress("UNUSED")
+class KaspressoRunner : AndroidJUnitRunner(), KaspressoRunNotifier by KaspressoRunNotifierImpl() {
+
+    override fun onCreate(arguments: Bundle) {
+        arguments.putArgs("listener", SpyRunListener::class.java.name)
```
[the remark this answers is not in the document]
Nikitae57: Done

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/runner/listener/SpyRunListener.kt
```diff
@@ -0,0 +1,40 @@
+package com.kaspersky.kaspresso.runner.listener
+
+import androidx.test.platform.app.InstrumentationRegistry
+import org.junit.runner.Description
+import org.junit.runner.Result
+import org.junit.runner.notification.Failure
+import org.junit.runner.notification.RunListener
+
+internal class SpyRunListener : RunListener() {
```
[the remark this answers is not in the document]
Nikitae57: Done

allure-support/src/main/kotlin/com/kaspersky/components/alluresupport/runlisteners/AllureRunListener.kt
```diff
@@ -0,0 +1,62 @@
...
+import io.qameta.allure.android.listeners.ExternalStoragePermissionsListener
+import io.qameta.allure.android.writer.TestStorageResultsWriter
+import io.qameta.allure.kotlin.Allure
+import io.qameta.allure.kotlin.junit4.AllureJunit4
+import io.qameta.allure.kotlin.util.PropertiesUtils
+import org.junit.runner.Description
+import org.junit.runner.Result
+import org.junit.runner.notification.Failure
+
+class AllureRunListener : KaspressoLateRunListener {
```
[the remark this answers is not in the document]
Nikitae57: Done

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/runner/listener/KaspressoLateRunListener.kt
Nikitae57: Done

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/runner/listener/KaspressoLateRunListener.kt
Nikitae57: Done

AzamatCherchesov (approved):

AzamatCherchesov (approved):

VladislavSumin (approved):

Issue #370, opened 2022-07-06
Gradle is unable to download sources on version 1.4.1


AndreThlckr:
When using newer versions of the library, Gradle is unable to find and download the library sources (which allow you to read and debug the source code directly on the IDE).

## Expected Behavior

Projects with the Kaspresso dependency should be able to download sources.

## Actual Behavior

When trying do download sources, the following error appears:

```console
* What went wrong:
Execution failed for task ':app:DownloadSources'.
> Could not resolve all files for configuration ':app:downloadSources_10c6f7e9-408b-4f6a-8bd9-fe15e255981e'.
   > Could not find com.kaspersky.android-components:kaspresso:1.4.1@aar.
     Searched in the following locations:
       - https://dl.google.com/dl/android/maven2/com/kaspersky/android-components/kaspresso/1.4.1@aar/kaspresso-1.4.1@aar.pom
       - https://repo.maven.apache.org/maven2/com/kaspersky/android-components/kaspresso/1.4.1@aar/kaspresso-1.4.1@aar.pom
     Required by:
         project :app
```

## Steps to Reproduce the Problem

  1. Create an empty project;
  2. Add the dependency `androidTestImplementation "com.kaspersky.android-components:kaspresso:1.4.1"`;
  3. Create a test using classes from Kaspresso;
  4. Try to access the source (on IntelliJ IDE, `Ctrl+left click` a Kaspresso class name/method call);
  5. You will only be able to see the decompiled bytecode.

## Specifications

  - Library version: at least >= 1.4.1
  - IDE used: Android Studio

## Observations
  - I haven't tested on all versions, but sources were able to be downloaded at least up to version 1.2.1.

Nikitae57:
Hi, @AndreThlckr! Thank you for your submission. Issue confirmed. We'll work on resolving it shortly and give you an update as soon, as fix would be ready

AzamatCherchesov:
Hi, @AndreThlckr! The fix will be available in the next release. Thanks for the issue!

Nikitae57:
Hello, again @AndreThlckr! We fixed this issue in the 1.5.1 release. Please, feel free to re-open the issue if you think that your problem isn't resolved

PR #428, opened 2022-11-17
fix no kaspresso sources issue && simplify publishing code

VladislavSumin:
fix issue #370.
Before fix we publish only `java` (not `kotlin`) sources.
see https://developer.android.com/studio/publish-library


build-logic/publication/src/main/kotlin/com/kaspersky/kaspresso/publication/AndroidLibraryPublishExtension.kt
```diff
@@ -1,25 +0,0 @@
...
-@Suppress("UnnecessaryAbstractClass")
-abstract class AndroidLibraryPublishExtension @Inject constructor(objects: ObjectFactory) {
-
-    /**
-     * android library variant to be published
-     *
-     * default: release
-     */
-    @Suppress("UnstableApiUsage")
-    val variant: Property<String> = objects.property<String>().convention("release")
```
VladislavSumin, line 17: I'remove this option because
* with new api impossible to simple use `afterEvaluate` to solve property usage
* we never use this feature before

AzamatCherchesov (approved):

matzuk:
@RuslanMingaliev ?

matzuk (approved):

Issue #433, opened 2022-11-18
Cannot setup interceptors :(


michalorminski:
Hi,

Apologies for the stupid question, but what and how should I use to implement a solution for a following type of problem:
"If a dialog X appears in a random moment in a Kaspresso test, close it by click on a UiButton with and id 'dismiss'". If the dialog is not appearing, just quickly move on with the test.

I've read the docs and figured out the maybe interceptors (or Watchers as we are also using some regular UiAutomator in the code but thats a different story) could help. 

I cannot figure out though how to setup an interceptor that looks over the tests and tries to click "dismiss" if an UiObjectNotFound will appear in some random moment and ofc if the "dismiss" button exist.

I am also not sure if I understand the idea of interceptors/watchers right probably not. :/

Thx in advance for the help.

michalorminski:
any1? :(

Areyana:
Thx for issue, we know about system dialog interceptor problem, i'll check it. As for interceptors they probably work without any code from your side in a "automatic" way (as for dialog/flaky/autoscroll interceptors for example).

Areyana:
fixed in #460

michalorminski:
Hi @Areyana , thx for fixing but that was not quite the issue I witnessed. I moved on a little bit though. Now I'm stuck at this example:

`lass SomeTest {
    @Before
    fun setup() {
        KautomatorConfigurator { // Kautomator runtime
            intercept {
                onUiInteraction { // Intercepting calls on UiInteraction classes across whole runtime
                    onPerform { uiInteraction, uiAction -> // Intercept perform() call
                        testLogger.i("KautomatorIntercept", "interaction=$uiInteraction, action=$uiAction")
                    }
                }
            }
        }
    }

    @Test
    fun test() {
        MyScreen {
            intercept {
                onUiInteraction { // Intercepting calls on UiInteraction classes while in the context of MyScreen
                    onCheck { uiInteraction, uiAssert -> // Intercept check() call
                        testLogger.i("KautomatorIntercept", "interaction=$uiInteraction, assert=$uiAssert")
                    }
                }
            }

            myView {
                intercept { // Intercepting ViewInteraction calls on this individual view
                    onPerform(true) { uiInteraction, uiAction -> // Intercept perform() call and overriding the chain
                        // When performing actions on this view, Kautomator level interceptor will not be called
                        // and we have to manually call UI Automator now.
[...]

PR #452, opened 2022-12-08
Tech sanity tests

Areyana:
part of https://github.com/KasperskyLab/Kaspresso/issues/439

+ Old \Sgoogle.android\S can't cover all cases that we need, so i'm adding SystemSafetyPattern enum with multiple patterns, which anyone can extend.

+ Sanity tests that cover interceptors logic, except autoscroll compose that will be part of new compose features PR.

PR #460, opened 2022-12-21
TECH: fix system dialog provider

Areyana:
related to #433


Nikitae57 (approved):
Would be cool to add tests for this case if there're no any yet

Areyana:
> Would be cool to add tests for this case if there're no any yet

Will be in #452

Nikitae57 (approved):

PR #474, opened 2023-01-10
TECH: use supported video recording resolution

Nikitae57:
Some devices do not support native resolution for screen recording (👀 https://developer.android.com/studio/command-line/adb.html#screenrecord). You are limited by avc capabilities

### DIFF
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/video/recorder/VideoRecordingThread.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/video/recorder/VideoRecordingThread.kt
@@ -1,15 +1,23 @@
 package com.kaspersky.kaspresso.device.video.recorder
 
+import android.app.Instrumentation
+import android.content.Context
+import android.media.MediaCodecList
+import android.os.Build
+import android.view.WindowManager
+import androidx.annotation.RequiresApi
 import androidx.test.uiautomator.UiDevice
 import com.kaspersky.kaspresso.logger.UiTestLogger
 import com.kaspersky.kaspresso.params.VideoParams
 import java.io.File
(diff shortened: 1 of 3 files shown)


AzamatCherchesov (dismissed):

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/video/recorder/VideoRecordingThread.kt
```diff
@@ -18,9 +26,51 @@ class VideoRecordingThread(
...
+    private fun startVideoRecordingWithNativeResolution() {
         execShellCommand("screenrecord --bit-rate ${params.bitRate} --bugreport ${file.absolutePath}")
     }
 
+    @RequiresApi(Build.VERSION_CODES.LOLLIPOP)
+    private fun startVideoRecordingWithRespectToCodecCapabilities() {
+        try {
+            val codecInfo = MediaCodecList(MediaCodecList.ALL_CODECS).codecInfos
+                .filter { it.isEncoder }
+                .find { it.name.contains("h264") }!!
```
matzuk, line 45: It's better to throw more valuable exception here instead of usage of `!!`
Please check other places here where `!!` is used

matzuk (approved):

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/video/recorder/VideoRecordingThread.kt
Nikitae57: In case of NPE outer catch block you mentioned in the previous comment would catch it and follback to native resolution recording

matzuk (approved):

PR #477, opened 2023-01-11
TECH: added new github workflow; use kaspresso snapshot dependency if…

Nikitae57:
… property is set

We want to check whether the project compiles with the new kaspresso release. This PR adds a new github workflow which we would trigger manually using snapshot version before release. It assembles android tests with kaspresso, compose support and allure support being used as remote snapshot dependencies instead of the project ones. Snapshot versions usage is determined by the gradle property which we pass in the introduced workflow

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -12,8 +12,14 @@ androidXTest = "1.5.0"
 testOrchestrator = "1.4.2"
 lifecycle = "2.5.1"
 thirdPartyReport = "0.17.798"
+kaspressoSnapshotVersion = "1.5.1-SNAPSHOT"
 
 [libraries]
+# kaspresso snapshots for release checks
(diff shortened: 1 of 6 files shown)


gradle/libs.versions.toml
```diff
@@ -12,8 +12,14 @@ androidXTest = "1.5.0"
 testOrchestrator = "1.4.2"
 lifecycle = "2.5.1"
 thirdPartyReport = "0.17.798"
+kaspressoSnapshotVersion = "1.5.1-SNAPSHOT"
```
AzamatCherchesov, line 15: @Nikitae57 maybe we should set snapshot version through input fields like here https://github.com/KasperskyLab/Kaspresso/blob/master/.github/workflows/publication_maven.yml. What do you think?

AzamatCherchesov: if this new action will be triggered by manual workflow run, than we will not need to set needed version in another commit before run

AzamatCherchesov (approved):

PR #478, opened 2023-01-11
Revert "ISSUE-404: Removed storage rules for all tests"

Nikitae57:
This reverts commit d202682fc53e0deabcfe977ccf2fead808ce4545.


AzamatCherchesov:
It would be nice to say why it was done =) @Nikitae57

AzamatCherchesov (approved):

Nikitae57:
Reverting deletion of storage rules is needed because we need storage permissions to use Documents folder on API <= 30

PR #522, opened 2023-06-02
TECH: (Silinsky) add dark mode screenshots support

### LINKED ISSUE #502: Screenshot tests for dark\light modes
It would be cool if we'd be able to make screenshot tests with dark\light mode switching

ReCursia:
Closes #502

### DIFF
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/rule/ThemeRule.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/rule/ThemeRule.kt
@@ -1,0 +1,32 @@
+package com.kaspersky.kaspresso.docloc.rule
+
+import androidx.appcompat.app.AppCompatDelegate
+import com.kaspersky.kaspresso.logger.UiTestLogger
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+
+class ThemeRule internal constructor(
+    private val changeAppTheme: Boolean,
+    private val logger: UiTestLogger
+) : TestRule {
+
+
+    override fun apply(base: Statement, description: Description): Statement {
+        return object : Statement() {
+            override fun evaluate() {
+                logger.i("DocLoc: changeAppThemeRule started")
+                try {
+                    if (changeAppTheme) {
+                        AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_YES)
+                        base.evaluate()
+                        AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_NO)
+                        base.evaluate()
+                    }
+                } finally {
+                    AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM)
+                }
+            }
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/rule/ThemeRule.kt
```diff
@@ -0,0 +1,32 @@
+package com.kaspersky.kaspresso.docloc.rule
+
+import androidx.appcompat.app.AppCompatDelegate
+import com.kaspersky.kaspresso.logger.UiTestLogger
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+
+class ThemeRule internal constructor(
```
Nikitae57, line 9: I would narrow down the name to tell user that this rule is used to toggle dark mode, not the whole theme

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/rule/ThemeRule.kt
```diff
@@ -0,0 +1,32 @@
...
+    private val changeAppTheme: Boolean,
```
Nikitae57, line 10: Same for argument

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/rule/ThemeRule.kt
```diff
@@ -0,0 +1,32 @@
...
+                logger.i("DocLoc: changeAppThemeRule started")
+                try {
+                    if (changeAppTheme) {
+                        AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_YES)
+                        base.evaluate()
+                        AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_NO)
+                        base.evaluate()
+                    }
+                } finally {
+                    AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM)
```
Nikitae57, line 27: Would be better to save preferences before test and restore it after instead of setting "follow system" unconditionally

Nikitae57 (changes requested):

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/rule/ThemeRule.kt
ReCursia: I've changed

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/rule/ThemeRule.kt
ReCursia: i've changed

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/rule/ThemeRule.kt
ReCursia: I've changed to use UNSPECIFIED, this is default value and also used to "reset" chosen night mode in app

Nikitae57 (approved):

PR #544, opened 2023-07-18
ISSUE-543: add parsing CollapsingToolbarLayout title + sample

mfglushchenko:
- add CollapsingToolbarLayout titles to the list of localized strings
- add sample screen and test

### DIFF
--- a/samples/kautomator-sample-app-upgrade/build.gradle.kts
+++ b/samples/kautomator-sample-app-upgrade/build.gradle.kts
@@ -11,6 +11,7 @@ android {
 
 dependencies {
     implementation(libs.appcompat)
+    implementation(libs.material)
     implementation(libs.multidex)
 
     androidTestImplementation(projects.kaspresso)


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/activities/metadata/ActivityMetadata.kt
```diff
@@ -61,7 +62,25 @@ internal class ActivityMetadata(
                     v.width,
                     v.height
                 )
+            }.toMutableList()
+
+        val stringsFromToolbarLayouts = TreeIterables.depthFirstViewTraversal(decorView)
```
AzamatCherchesov, line 67: Can we avoid traversing the hierarchy twice? For example:
TreeIterables.depthFirstViewTraversal(decorView)
   .filter { it.visibility == View.VISIBLE }
   .filter( it is TextView || it is CollapsingToolbarLayout)
   .map { case for TextView and CollapsingToolbarLayout } ...

AzamatCherchesov (dismissed):

AzamatCherchesov:
Some problems with detekt. @mfglushchenko have a look pls

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/activities/metadata/ActivityMetadata.kt
mfglushchenko: done

AzamatCherchesov (approved):

PR #569, opened 2023-09-05
ISSUE-561: setup Kakao and Kautomator settings before each test

sumin93:
Fix https://github.com/KasperskyLab/Kaspresso/issues/561


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/testcases/core/TestRunner.kt
```diff
@@ -94,6 +103,28 @@ internal class TestRunner<InitData, Data>(
         resultException?.let { throw it }
     }
 
+    private fun injectKaspressoInKakao() {
```
matzuk, line 106: Maybe it makes sense to put this inject-specific code in a single place to avoid any confusion in the future.
Also, I see some code duplication here.

matzuk (approved):

PR #573, opened 2023-09-18
TECH: fix source sets

Nikitae57:


AzamatCherchesov:
Have we any automated checks for this docs?

Nikitae57:
> Have we any automated checks for this docs?

Not yet

PR #597, opened 2023-11-16
ISSUE-576: Allow complex ADB server commands execution

### LINKED ISSUE #576: Enable execution of complex shell commands via AdbServer
Sometimes we use AdbServer for triggering an execution of shell commands on the PC, however because `CmdCommandPerformer::perform()` uses `Runtime::exec(String command)` instead of  `Runtime::exec(String[] cmdarray)` it is impossible to execute commands which contain whitespaces in their arguments, or commands which use output redirection or pipes. 

It would be great if it was possible to execute commands via `Runtime::exec(String[] cmdarray)`.
E.g. instead of command `adbServer.performCmd("script.sh \"some text\"")` which fails, execute command `adbServer.performCmd(arrayOf("script.sh", "some text")` which would execute correctly.
Another example is `adbServer.performCmd(arrayOf("sh", "-c",  "adb shell dumpsys deviceidle | grep mForceIdle")`

Nikitae57:
Closes #576


eakurnikov (dismissed):

adb-server/adb-server-command-types/src/main/java/com/kaspersky/adbserver/commandtypes/AdbCommand.kt
```diff
@@ -2,4 +2,4 @@ package com.kaspersky.adbserver.commandtypes
```
matzuk, line 2: General feedback.

I guess we can use `Runtime::exec(String[] cmdarray)` instead of `Runtime::exec(String command)` in `CmdCommandPerformer.perform`. Have a look at `java.lang.Runtime`, where both methods call the same method and `command` param in `Runtime::exec(String command)` is parsing into the array of strings.

I don't support introducing separate classes like `ComplexAdbCommand`. I think we need to add the new constructors like `Command(command: String, arguments: List<String>)` and the correspondent methods in `AdbTerminal` and etc.

I don't support syntaxis like `(command: List<String>)` and `(command): String`. It's confusing. Also, the most popular syntaxis in these cases is `(command: String, arguments: List<String>)`. Yes, another command is an argument of the main command. Cases like `sh -c "adb shell dumpsys deviceidle | grep mForceIdle"` should be described in the documentation and examples.

matzuk (changes requested):

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/server/AdbServerImpl.kt
```diff
@@ -95,16 +73,27 @@ class AdbServerImpl(
...
-                if (result.status == ExecutorResultStatus.TIMEOUT) {
-                    throw AdbServerException(
-                        """
+            .onEach { (command, result) -> logCommandResult(command, result) }
+            .map { (_, result) -> result.description }
+            .toList()
+    }
+
+    private fun performComplex(command: String, arguments: List<String>, executor: (String, List<String>) -> CommandResult): String {
+        logger.i("AdbServer. The complex command to execute=$command")
```
matzuk, line 82: `arguments` missed

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/server/AdbServerImpl.kt
```diff
@@ -30,58 +30,36 @@ class AdbServerImpl(
...
             adbTerminal.executeAdb("shell $it")
         }
     }
 
-    /**
-     * Disconnect from AdbServer.
-     * The method is called by Kaspresso after each test.
-     */
+    override fun performCmd(command: String, arguments: List<String>): String {
+        return performComplex(command, arguments, adbTerminal::executeCmd)
```
matzuk, line 52: I guess it's better to use the single `perform` method. Don't see why we need `performComplex` method.

docs/Wiki/Executing_adb_commands.en.md
```diff
@@ -190,6 +190,14 @@ Now the logs looks like:
 2020-09-10 12:24:27.427 10349-10378/com.kaspersky.kaspressample I/KASPRESSO_ADBSERVER: The result of command=AdbCommand(body=shell su 0 svc data disable) => CommandResult(status=SUCCESS, description=exitCode=0, message=, serviceInfo=The command was executed on desktop=Desktop-30548)
 ```
 
+## Complex commands
```
matzuk, line 193: The description is unclear for people who don't have the task context. 
First of all, provide examples to show the difference between the two methods.

Another point is the existence of two methods that are similar to each other but their implementations are different in terms of tokenization and etc, is not great. This case may confuse devs. I think we should deprecate the first method with a very detailed explanation. I doubt that we will remove the first method but we'll give a clear picture for devs about which method they need to use.

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/server/AdbServer.kt
```diff
@@ -26,6 +26,20 @@ interface AdbServer {
      */
     fun performCmd(vararg commands: String): List<String>
 
+    /**
+     * Performs shell commands blocking current thread. Allows more control over how arguments are parsed.
```
matzuk, line 30: Enhance the comment section following my suggestions above.

matzuk (changes requested):

adb-server/adb-server-device/src/main/java/com/kaspersky/adbserver/device/AdbTerminal.kt
```diff
@@ -27,10 +27,32 @@ object AdbTerminal {
         AdbCommand(command)
     ) ?: throw IllegalStateException("Please first of all call [connect] method to establish a connection")
 
+    /**
+     * Allows more control over how arguments are parsed. Each element in the [arguments] list
+     * is used as is without tokenizing.
+     * Refer to the Runtime::exec(String[] cmdarray)
```
matzuk, line 33: Please provide a full link to this method.

adb-server/adb-server-device/src/main/java/com/kaspersky/adbserver/device/AdbTerminal.kt
```diff
@@ -27,10 +27,32 @@ object AdbTerminal {
...
      * Please first of all call [connect] method to establish a connection
      */
     fun executeCmd(command: String): CommandResult = device?.fulfill(
         CmdCommand(command)
     ) ?: throw IllegalStateException("Please first of all call [connect] method to establish a connection")
+
+    /**
+     * Allows more control over how arguments are parsed. Each element in the [arguments] list
+     * is used as is without tokenizing.
+     * Refer to the Runtime::exec(String[] cmdarray)
```
matzuk, line 51: Please provide a full link to this method.

docs/Wiki/Executing_adb_commands.en.md
```diff
@@ -50,6 +54,23 @@ fun test() =
         // ....
 }
 ```
+This method passes each command to the java runtime which tokenizes it by whitespaces. It could be not ideal. It can't be used for the commands with
+the whitespaces in their arguments (e.g. `adb pull "path/with whitespace/file"`) and it doesn't allow piping (e.g. `cat $BUILD_DIR/report.txt | grep filte > filtered.txt`).
```
matzuk, line 58: "path/with whitespace/file" is a bit confusing.
maybe "path_with whitespase to_file" is better? but it is up to you.

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/server/AdbServer.kt
```diff
@@ -24,8 +24,29 @@ interface AdbServer {
      * @throws AdbServerException if a result status of any command in @param commands is Failed
      * @return list of answers of commands' execution
      */
+    @Deprecated("This method doesn't work for the commands with the complex arguments containing " +
+            "whitespaces (e.g. 'adb pull \"path/with whitespace/file\") and doesn't allow commands piping like" +
+            "adbServer.performCmd(\"bash\", listOf(\"-c\", \"adb shell dumpsys deviceidle | grep mForceIdle\"))" +
+            "which the AdbServer.performCmd(String, List<String>) does. " +
+            "For more details, please check out https://github.com/KasperskyLab/Kaspresso/blob/master/docs/Wiki/Executing_adb_commands.en.md",
+        ReplaceWith("adbServer.performCmd(*commands, emptyList())")
```
matzuk, line 32: Are you sure that it is correct replacement? You put an array (`*commands`) where `adbServer.performCmd` expects `command: String`.

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/server/AdbServer.kt
```diff
@@ -24,8 +24,29 @@ interface AdbServer {
...
+    )
     fun performCmd(vararg commands: String): List<String>
 
+    /**
+     * Performs shell commands blocking current thread. Allows more control over how arguments are parsed.
+     * Each list element is used as is. Refer to the https://docs.oracle.com/javase/8/docs/api/java/lang/ProcessBuilder.html.
```
matzuk, line 38: Please unify your comments. Somewhere you write class and method name, somewhere you provide a link.

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/server/AdbServer.kt
```diff
@@ -36,8 +57,29 @@ interface AdbServer {
      * @throws AdbServerException if a result status of any command in @param commands is Failed
      * @return list of answers of commands' execution
      */
+    @Deprecated("This method doesn't work for the commands with the complex arguments containing " +
+            "whitespaces (e.g. 'adb pull \"path/with whitespace/file\") and doesn't allow commands piping like" +
+            "adbServer.performCmd(\"bash\", listOf(\"-c\", \"adb shell dumpsys deviceidle | grep mForceIdle\"))" +
+            "which the AdbServer.performCmd(String, List<String>) does " +
+            "For more details, please check out https://github.com/KasperskyLab/Kaspresso/blob/master/docs/Wiki/Executing_adb_commands.en.md",
+        ReplaceWith("adbServer.performAdb(*commands, emptyList())")
```
matzuk, line 65: Again, I am not sure about your replacement.

adb-server/adb-server-command-types/src/main/java/com/kaspersky/adbserver/commandtypes/AdbCommand.kt
```diff
@@ -2,4 +2,7 @@ package com.kaspersky.adbserver.commandtypes
 
 import com.kaspersky.adbserver.common.api.Command
```
matzuk, line 3: Could you also reflect all these cases in tests? I mean testing a single command, a command with arguments and a command with commands in arguments (complex commands with pipes and etc.).

matzuk (changes requested):

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/server/AdbServer.kt
Nikitae57: It's the best from the worst. If you leave just `commands` then IDE will replace `adbServer.performCmd("abc")` with `adbServer.performCmd(arrayOf("abc"), emptyList())` which will be highlighted as an error. I think that 90% of adbServer.perform... calls are with a signle command. For those cases, when there're several commands, then it will be replaced with `adbServer.performCmd("asd", "abc", emptyList())` which is ugly too, but less possible. It's a tradeoff. I don't want to leave ReplaceWith statement with a stub, because it will upset users by erasing their precious commands

matzuk (dismissed):

AzamatCherchesov (approved):

PR #598, opened 2023-11-21
TECH: Draft marathon config

Nikitae57:


.github/workflows/marathon_33.yml
```diff
@@ -0,0 +1,29 @@
...
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+      - name: Assemble android tests
+        run: ./gradlew -PCI=true assembleDebugAndroidTest assembleDebug
+      - name: Print
+        run: find -name "*.apk"
+      - name: run tests on android 33
+        uses: MarathonLabs/action-test@0.3.3
+        with:
+          apiKey: ${{ secrets.marathon_github_token }}
```
Vacxe, line 25: Just FYI, secrets except `GITHUB_TOKEN` is not accessible for `pull_request` in case of PRs from forks

PR #643, opened 2024-05-15
TECH: pass SystemLanguage instance through Device class

Nikitae57:


samples/kaspresso-sample/src/androidTest/kotlin/com/kaspersky/kaspressample/systemlanguage/ChangeSystemLanguageMidTestCase.kt
```diff
@@ -24,20 +23,19 @@ class ChangeSystemLanguageMidTestCase : TestCase() {
 
     @Test
     fun checkLocaleChangeMidTest() = run {
-        val systemLanguage = SystemLanguage(device.targetContext, testLogger, device.hackPermissions)
```
AzamatCherchesov, line 27: maybe we should mark SystemLanguage as Deprecated or change its visibility?

AzamatCherchesov (approved):

PR #655, opened 2024-07-24
TECH: Screenshot comparison tests

Nikitae57:


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/files/resources/impl/DefaultResourcesRootDirsProvider.kt
```diff
@@ -6,6 +6,8 @@ import java.io.File
 class DefaultResourcesRootDirsProvider : ResourcesRootDirsProvider {
     override val logcatRootDir = File("logcat")
     override val screenshotsRootDir = File("screenshots")
+    override val originalScreenshotsRootDir = File("original_screenshots")
+    override val screenshotsDiffRootDir = File("diff")
```
CherchesovAzamat, line 10: It might be better to rename this diff file. The context of the screenshots is lost

kaspresso/gradle.properties
```diff
@@ -1,4 +1,5 @@
 publish.artifactGroup=com.kaspersky.android-components
 publish.artifactName=kaspresso
 publish.publicationName=Kaspresso
-publish.bintrayRepo=Kaspresso
\ No newline at end of file
+publish.bintrayRepo=Kaspresso
```
CherchesovAzamat, line 4: let's move from root folder to sample dir

AzamatCherchesov (approved):

PR #665, opened 2024-08-23
Добавил зависимости в build.gradle

Xqluziff:


Nikitae57:
Declaring the dependencies is a step in the tutorial. We don't want to skip it
https://kasperskylab.github.io/Kaspresso/ru/Tutorial/Writing_simple_test/#_2

PR #683, opened 2024-11-13
ISSUE-675: sanitize filename in video recorder

ersanin:


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/video/VideosImpl.kt
```diff
@@ -12,7 +12,10 @@ class VideosImpl(
 ) : Videos {
 
     override fun record(tag: String) {
-        val videoFile: File = resourceFilesProvider.provideVideoFile(tag)
+        val sanitizedVideoName = tag
```
CherchesovAzamat, line 15: Will this be a breaking change or not?

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/video/VideosImpl.kt
```diff
@@ -12,7 +12,10 @@ class VideosImpl(
...
+            .replace("[/:\"*?<>|]+".toRegex(), "")
```
Nikitae57, line 16: 1. The issue creator used backing slash to cover each special simbol. It's not needed?
2. Please, replace with the undescrore. User expected a simbol there
3. Please, if the video name changed, then leave a warning log with something like `log("Can't record video with name $tag since it contains one of the following special characters [/:\"*?<>|]. Changing the name to $sanitizedVideoName")`

Nikitae57 (changes requested):
Minor issues. May release as is, but I would change it in the future

Nikitae57 (approved):

PR #700, opened 2025-01-28
Issue 699 compose doc loc metadata

### LINKED ISSUE #399: Add screenshots for compose @Preview
Add ability to automatically screenshots all @Preview in compose

Nikitae57:
Closes #399

### DIFF
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/DocLocScreenshotCapturer.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/DocLocScreenshotCapturer.kt
@@ -1,6 +1,7 @@
 package com.kaspersky.kaspresso.docloc
 
 import com.kaspersky.kaspresso.device.screenshots.screenshotmaker.ScreenshotMaker
+import com.kaspersky.kaspresso.docloc.metadata.saver.MetadataSaver
 import com.kaspersky.kaspresso.files.resources.ResourceFilesProvider
 import com.kaspersky.kaspresso.internal.wait.wait
 import com.kaspersky.kaspresso.logger.UiTestLogger
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/metadata/extractor/UiMetadataExtractor.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/metadata/extractor/UiMetadataExtractor.kt
@@ -1,0 +1,55 @@
+package com.kaspersky.kaspresso.docloc.metadata.extractor
+
+import android.app.Activity
+import androidx.test.uiautomator.By
+import androidx.test.uiautomator.UiDevice
+import com.kaspersky.kaspresso.device.activities.Activities
+import com.kaspersky.kaspresso.docloc.metadata.LocalizedString
+import com.kaspersky.kaspresso.docloc.metadata.Metadata
+import com.kaspersky.kaspresso.docloc.metadata.Window
+
+internal class UiMetadataExtractor(
+    private val uiDevice: UiDevice,
+    private val activities: Activities,
+) : MetadataExtractor {
+
+    private val metadataExtractorHelper = MetadataExtractorHelper()
+
+    override fun getMetadata(): Metadata {
+        val activity = activities.getResumed() ?: throw RuntimeException("Failed to get current activity")
+        return createMetadata(activity)
+    }
+
+    private fun createMetadata(activity: Activity): Metadata {
+        val localizedStrings = metadataExtractorHelper.resolveAmbiguous(
+            getLocalizedStrings()
+        )
+        val window = activity.window.decorView.run {
+            Window(
+                left,
+                top,
+                width,
+                height,
+                localizedStrings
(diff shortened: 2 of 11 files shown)


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/metadata/extractor/UiMetadataExtractor.kt
```diff
@@ -0,0 +1,96 @@
...
+                height,
+                localizedStrings
+            )
+        }
+        return Metadata(window)
+    }
+
+    private fun getLocalizedStrings(): List<LocalizedString> {
+        uiDevice.setCompressedLayoutHierarchy(false)
+        val objectsWithText = uiDevice.findObjects(By.enabled(true))
```
Nikitae57, line 44: 100% needs better UiObjects tree parsing. Probably would be better if we used recursion

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/DocLocScreenshotCapturer.kt
```diff
@@ -1,6 +1,7 @@
 package com.kaspersky.kaspresso.docloc
 
 import com.kaspersky.kaspresso.device.screenshots.screenshotmaker.ScreenshotMaker
+import com.kaspersky.kaspresso.docloc.metadata.saver.MetadataSaver
```
AzamatCherchesov, line 4: only import without usage?

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/metadata/extractor/UiMetadataExtractor.kt
```diff
@@ -0,0 +1,96 @@
...
+                    logger.e("UiMetadataExtractor::getLocalizedStrings() - Error while loading object")
+                    false
+                }
+            }
+
+        val localizedStrings = mutableListOf<LocalizedString>()
+        for (obj in objectsWithText) {
+            try {
+                val resName = obj.resourceName ?: continue
+                if (!obj.text.isNullOrBlank()) {
```
AzamatCherchesov, line 60: if you have else block you do not need !

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/metadata/extractor/UiMetadataExtractor.kt
```diff
@@ -0,0 +1,96 @@
...
+                } catch (ex: StaleObjectException) {
+                    logger.e("UiMetadataExtractor::getLocalizedStrings() - Error while loading object")
+                    false
+                }
+            }
+
+        val localizedStrings = mutableListOf<LocalizedString>()
+        for (obj in objectsWithText) {
+            try {
+                val resName = obj.resourceName ?: continue
```
AzamatCherchesov, line 59: maybe it will be better to throw exception? Or we really need to skip item without res name?

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/metadata/extractor/UiMetadataExtractor.kt
```diff
@@ -0,0 +1,96 @@
...
+                            top = coords.top,
+                            width = coords.width(),
+                            height = coords.height(),
+                        )
+                    )
+                } else {
+                    obj.children
+                        .filter { !it.text.isNullOrBlank() }
+                        .forEach {
+                            val coords = obj.visibleBounds
```
AzamatCherchesov, line 76: same code can be moved to a separate method

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/DocLocScreenshotCapturer.kt
Nikitae57: It is used in the consctructor. Before my changes MetadataSaver was in the same pachage as DocLocScreenshotCapturer. I made renamed it to the DefautMetadataSaver and extracted the interface MetadataSaver. Now it has to be imported

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/metadata/extractor/UiMetadataExtractor.kt
Nikitae57: The idea is to pick the view that might be the text container or a text itself. Either way we need an id which might be nullable, but I don't think we have to throw an exception

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/docloc/metadata/extractor/UiMetadataExtractor.kt
Nikitae57: You found a possible bug. When we iterate through the children we better use the child coordinates

AzamatCherchesov (approved):

PR #718, opened 2025-04-28
ISSUE-717: Add "don't ask again" Deny button in Permissions

### LINKED ISSUE #717: Permissions.denyViaDialog() only works once
**Describe the bug**
A user can click the **Deny** button in the permissions dialog to dismiss it without granting the permission. But as of Android 11, if the app asks about the permission again, the **Deny** button will have a different ID, one that Kaspresso currently doesn't know about, so trying to press it again fails.

**To Reproduce**
1. The app should ask for a runtime permission.
2. Call `denyViaDialog()`.
3. The app should ask for the same permission again.
4. Try to call `denyViaDialog()` again.

**Expected behavior**
The **Deny** button gets clicked both times.

**Actual behavior**
The second click of the **Deny** button never happens.

https://github.com/user-attachments/assets/a6e40e90-a3b3-4d7b-a488-cebd84de3fce

**Smartphone (please complete the following information):**
 - Device: Google Pixel 7
 - OS: Android 15

**Additional context**
#667 added an option to specify a different **Allow** button than the default. This issue can easily be fixed by adding a button with the resource ID `permission_deny_and_dont_ask_again_button` and an optional parameter to `denyViaDialog()`.

JordanLongstaff:
Fixes #717

Arguably, `allowViaDialog` and `denyViaDialog` are rather redundant, since they just do the same thing as `clickButton`, only providing default arguments.

### DIFF
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/permissions/Permissions.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/permissions/Permissions.kt
@@ -16,7 +16,7 @@ interface Permissions {
     /**
      * Passes the permission-requesting permissions dialog and denies permissions.
      */
-    fun denyViaDialog()
+    fun denyViaDialog(button: Button = Button.DENY)
 
     /**
      * Check the permission-requesting permissions dialog is visible.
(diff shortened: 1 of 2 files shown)


JordanLongstaff:
Bump. I have to use UiAutomator myself as a workaround. Kaspresso should implement this ASAP.

PR #719, opened 2025-04-28
TECH: Update compose

Nikitae57:


Nikitae57:
This update makes us raise min SDK to 24

gradle/libs.versions.toml
```diff
@@ -1,13 +1,13 @@
 [versions]
-kotlin = "1.8.22"
+kotlin = "2.0.0"
```
javadev, line 2: 👍

PR #722, opened 2025-06-19
Update EnvironmentDetector.kt

GeorgCantor:


touhidurrr (commented):
isn't `runCatching` heavier instead
this account was opening pr's with llms. so I was tracking its past history. and see what it does.

PR #734, opened 2026-02-26
ISSUE-733: Add configurable screenshotMaker to Kaspresso.Builder

### LINKED ISSUE #733: Add configurable screenshotMaker to Kaspresso.Builder
## Problem

`InternalScreenshotMaker` uses `Activity.window.decorView.draw(Canvas)` which only captures the main activity window. Views rendered in separate popup windows — such as Material 3's `ModalBottomSheet` — are invisible in the resulting screenshot.

`CombinedScreenshotMaker` falls back to `ExternalScreenshotMaker` (which uses `UiDevice.takeScreenshot()` and captures the full screen) only when the internal one throws an exception:

```kotlin
runCatching {
    preferredScreenshotMaker.takeScreenshot(file)
}.onFailure {
    fallbackScreenshotMaker.takeScreenshot(file)
}
```

But `InternalScreenshotMaker` doesn't throw — it succeeds and produces an incomplete image. So the fallback never runs.

## Current workaround

There is no clean way to swap the screenshot maker. The `screenshotMaker` is constructed inline inside `postInitVariables()` and not exposed as a configurable property.

The only working workaround is to avoid `advanced()` entirely, use `simple()`, manually reconstruct `ScreenshotsImpl` with all its dependencies in `.apply`, and re-add interceptors by hand:

```kotlin
val E2ETestsBuilder = Kaspresso.Builder.simple {
    flakySafetyParams = FlakySafetyParams.custom(timeoutMs = 1_000)
}.apply {
    screenshots = ScreenshotsImpl(
        logger = libLogger,
        resourceFilesProvider = resourceFilesProvider,
        screenshotMaker = ExternalScreenshotMaker(
            instrumentalDependencyProviderFactory.getComponentProvider<ExternalScreenshotMaker>(
                instrumentation
            ),
            screenshotParams
        ),
        visualTestParams = visualTestParams,
        screenshotsComparator = screenshotsComparator,
        dirsProvider = dirsProvider,
        resourceFileNamesProvider = resourceFileNamesProvider,
        resourcesDirsProvider = resourcesDirsProvider,
    )
    stepWatcherInterceptors.add(ScreenshotStepWatcherInterceptor(screenshots))
    testRunWatcherInterceptors.add(VideoRecordingInterceptor(videos))
    stepParams.autonumber = true
}
```

This is fragile — it depends on Builder internals, and using `advanced()` or `withForcedAllureSupport()` instead of `simple()` doesn't work because those builders wire screenshot interceptors before `.apply` runs, capturing a stale reference to the old `screenshots` object.

## Proposed fix

1. Add a `lateinit var screenshotMaker: ScreenshotMaker` property to `Kaspresso.Builder`, following the same `if (!::prop.isInitialized)` pattern used by other Builder properties. The default behavior is unchanged — if not set, the existing `CombinedScreenshotMaker` is created.

2. Add a secondary constructor to `ExternalScreenshotMaker` that accepts `Instrumentation` directly, so users don't need to deal with `InstrumentalDependencyProvider`.

3. Fix `ExternalScreenshotMaker.takeFullWindowScreenshot` which currently throws `TODO("not yet implemented")`. Since `UiDevice.takeScreenshot()` already captures the full screen, it can delegate to `takeScreenshot()`.

This reduces the workaround above to a single line that works with any builder — `simple()`, `advanced()`, or `withForcedAllureSupport()`:

```kotlin
Kaspresso.Builder.advanced {
    screenshotMaker = ExternalScreenshotMaker(instrumentation)
}
```

Twinsen81:
Closes #733

## Summary

- Add `lateinit var screenshotMaker: ScreenshotMaker` to `Kaspresso.Builder`, following the same `if (!::prop.isInitialized)` pattern used by other Builder properties. Default behavior is unchanged.
- Add a secondary constructor to `ExternalScreenshotMaker` that accepts `Instrumentation` directly, so users don't need to deal with `InstrumentalDependencyProvider`.
- Fix `ExternalScreenshotMaker.takeFullWindowScreenshot` which previously threw `TODO("not yet implemented")`. Since `UiDevice.takeScreenshot()` already captures the full screen, it delegates to `takeScreenshot()`.

## Usage

```kotlin
Kaspresso.Builder.advanced {
    screenshotMaker = ExternalScreenshotMaker(instrumentation)
}
```

## Test plan

- [ ] Verify default behavior is unchanged: `Kaspresso.Builder.simple {}` and `advanced {}` without setting `screenshotMaker` should produce the same `CombinedScreenshotMaker` as before
- [ ] Verify custom screenshotMaker: setting `screenshotMaker = ExternalScreenshotMaker(instrumentation)` in `simple {}` or `advanced {}` uses UiDevice-based capture
- [ ] Verify popup windows (e.g. ModalBottomSheet) are visible in screenshots when using `ExternalScreenshotMaker`
- [ ] Verify `takeFullWindowScreenshot` no longer throws on `ExternalScreenshotMaker`

### DIFF
--- a/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/screenshots/screenshotmaker/ExternalScreenshotMaker.kt
+++ b/kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/screenshots/screenshotmaker/ExternalScreenshotMaker.kt
@@ -1,6 +1,5 @@
 package com.kaspersky.kaspresso.device.screenshots.screenshotmaker
 
-import android.app.Instrumentation
 import androidx.test.uiautomator.UiDevice
 import com.kaspersky.kaspresso.instrumental.InstrumentalDependencyProvider
 import com.kaspersky.kaspresso.params.ScreenshotParams
@@ -19,19 +18,28 @@ class ExternalScreenshotMaker(
 ) : ScreenshotMaker {
 
     /**
-     * Creates an [ExternalScreenshotMaker] from an [Instrumentation] instance.
+     * Creates an [ExternalScreenshotMaker] from a [UiDevice] instance.
      *
      * This is a convenience constructor for use in [com.kaspersky.kaspresso.kaspresso.Kaspresso.Builder]:
      * ```
      * Kaspresso.Builder.simple {
-     *     screenshotMaker = ExternalScreenshotMaker(instrumentation)
+     *     screenshotMaker = ExternalScreenshotMaker(UiDevice.getInstance(instrumentation))
      * }
      * ```
      */
     constructor(
-        instrumentation: Instrumentation,
+        uiDevice: UiDevice,
         params: ScreenshotParams = ScreenshotParams()
-    ) : this(SimpleUiDeviceProvider(instrumentation), params)
+    ) : this(
+        object : InstrumentalDependencyProvider {
+            override val isAndroidRuntime = true
+            override val uiDevice get() = uiDevice
+            override val uiAutomation get() = throw UnsupportedOperationException()
+            override val runNotifier get() = throw UnsupportedOperationException()
+            override fun getUiAutomation(flags: Int) = throw UnsupportedOperationException()
+        },
+        params
+    )
 
     private val device: UiDevice
         get() = instrumentalDependencyProvider.uiDevice
@@ -45,18 +53,4 @@ class ExternalScreenshotMaker(
     }
 
     override fun takeFullWindowScreenshot(file: File) = takeScreenshot(file)
-
-    /**
-     * Minimal [InstrumentalDependencyProvider] that only provides [UiDevice].
-     * Used by the [Instrumentation]-based constructor to avoid depending on the full provider implementation.
-     */
-    private class SimpleUiDeviceProvider(
-        private val instrumentation: Instrumentation
-    ) : InstrumentalDependencyProvider {
-        override val isAndroidRuntime: Boolean = true
-        override val uiDevice: UiDevice get() = UiDevice.getInstance(instrumentation)
-        override val uiAutomation get() = instrumentation.uiAutomation
-        override val runNotifier get() = throw UnsupportedOperationException()
-        override fun getUiAutomation(flags: Int) = instrumentation.getUiAutomation(flags)
-    }
 }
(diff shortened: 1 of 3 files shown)


kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/screenshots/screenshotmaker/ExternalScreenshotMaker.kt
```diff
@@ -24,7 +44,19 @@ class ExternalScreenshotMaker(
...
-        TODO("External Full Window Screenshot not yet implemented")
+    override fun takeFullWindowScreenshot(file: File) = takeScreenshot(file)
+
+    /**
+     * Minimal [InstrumentalDependencyProvider] that only provides [UiDevice].
+     * Used by the [Instrumentation]-based constructor to avoid depending on the full provider implementation.
+     */
+    private class SimpleUiDeviceProvider(
+        private val instrumentation: Instrumentation
+    ) : InstrumentalDependencyProvider {
```
Nikitae57: TBH, looks unnecessary. Maybe, create another constructor with just uiDeviсe? If the user has an instrumentation, then he could construct uiDeivce in place with UiDevice.getInstance(instrumentation)

Nikitae57 (changes requested):
Overall looks great. Thank you! SimpleUiDeviceProvider needs a second look though

kaspresso/src/main/kotlin/com/kaspersky/kaspresso/device/screenshots/screenshotmaker/ExternalScreenshotMaker.kt
Twinsen81: Makes sense. I can replace it with an anonymous object:
```
    constructor(
        uiDevice: UiDevice,
        params: ScreenshotParams = ScreenshotParams()
    ) : this(
        object : InstrumentalDependencyProvider {
            override val isAndroidRuntime = true
            override val uiDevice get() = uiDevice
            override val uiAutomation get() = throw UnsupportedOperationException()
            override val runNotifier get() = throw UnsupportedOperationException()
            override fun getUiAutomation(flags: Int) = throw UnsupportedOperationException()
        },
        params
    )
```
And then on the call site, it'll be:
```
Kaspresso.Builder.simple {
    screenshotMaker = ExternalScreenshotMaker(UiDevice.getInstance(instrumentation))
}
```

Does it look better?

Nikitae57: Looks good

Nikitae57 (approved):

